PR #4, opened 2015-03-06
Use jinja2 and pyramid_jinja2

elemoine:
As discussed we'll use jinja2 as the template engine/language. This commit removes the pyramid_mako dependency and adds the pyramid_jinja2 dependency. The pyramid_chameleon dependency will be removed in the future. It's not removed for now as the default page relies on it.


ingenieroariel:
Awesome. +1

PR #8, opened 2015-03-13
Add admin unit search web service

elemoine:
This PR adds a JSON web service for searching admin units, and make the search field rely on that search service.

The URL of the service is `/adminunit` and the `q` parameter is used for searching admin units. For example: `/adminunit?q=mor`.

### DIFF
--- a/thinkhazard/models.py
+++ b/thinkhazard/models.py
@@ -1,11 +1,10 @@
 from sqlalchemy import (
     Column,
-    Index,
     Integer,
-    Text,
+    Unicode,
     )
 
-from sqlalchemy.ext.declarative import declarative_base
+from sqlalchemy.ext.declarative import declarative_base, AbstractConcreteBase
 
 from sqlalchemy.orm import (
     scoped_session,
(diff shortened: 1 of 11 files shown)


thinkhazard/models.py
```diff
@@ -18,10 +17,50 @@
...
+
+    def __json__(self, request):
+        return {'admin0': self.admin0_name, 'admin1': self.name}
+
+
+class Admin2(Admin):
+    __tablename__ = 'admin2'
+    order = 2
+
+    admin0_name = Column(Unicode)
```
ingenieroariel: I agree this denormalization (admin0 field on admin2) is a good idea for performance reasons. Is there a need for a constraint or a trigger at the database level to ensure it stays in sync?

elemoine: The input data was already denormalized in that respect.

fvanderbiest: We just took the source data "as they are".

If we were to do joins, we would need working foreign keys, hence uniqueness on GAUL codes ... which we do not have ATM.

ingenieroariel:
:+1:

elemoine:
@ingenieroariel We're going to update the demo server, and send a link to the group.

PR #9, opened 2015-03-13
Make debug and node_modules renderer globals

elemoine:
As they will be used in every page of the application.


fvanderbiest:
Looks good.

PR #12, opened 2015-03-27
GeoServer "data dir"

fvanderbiest:
This PR adds a minimal "data dir" for GeoServer. It stems directly from https://github.com/geoserver/geoserver/tree/2.7.x/data/minimal
 
This (ill-named) "data dir" will be updated when data is available to create store and layer configurations.


elemoine:
Looks good. I just think that `geoserver_datadir` would make a more explicit merge.

PR #13, opened 2015-03-27
Add Pyramid's cache buster

elemoine:
With this PR we use Pyramid's cache buster for the application's static resources. This is to avoid problems because of the Varnish cache we use on AWS.

This requires using the development version of Pyramid (1.6.dev0), which we download from GitHub.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,12 +25,8 @@ help:
 	@echo
 
 .PHONY: install
-install: setup-develop .build/node_modules.timestamp
+install: .build/requirements.timestamp .build/node_modules.timestamp
 	
-.PHONY: setup-develop
-setup-develop: .build/venv
-	.build/venv/bin/python setup.py develop
-
 .PHONY: build
 build: buildcss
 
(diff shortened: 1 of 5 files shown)


fvanderbiest:
LGTM

Makefile
```diff
@@ -134,6 +135,15 @@ thinkhazard/static/build/%.css: $(LESS_FILES) .build/node_modules.timestamp
 	sed -e 's#{{PYTHONPATH}}#$(shell .build/venv/bin/python -c "import distutils; print(distutils.sysconfig.get_python_lib())")#' \
 		-e 's#{{WSGISCRIPT}}#$(abspath .build/venv/thinkhazard.wsgi)#' $< > $@
 
+.build/pyramid.zip:
+	mkdir -p $(dir $@)
+	wget -O $@ https://github.com/Pylons/pyramid/archive/master.zip
```
ingenieroariel: Is it safe to get pyramid from master instead of a tag?

ingenieroariel: I looked at the pyramid repo and there is not tag for that development version. I guess we just need to keep an eye out for the next release and update this link and the setup.py to the released artifact.

ingenieroariel:
:+1:

Makefile
elemoine: > Is it safe to get pyramid from master instead of a tag?

You're right, it would be better to use a fixed version. I'll change it before merging. Thanks.

PR #17, opened 2015-04-16
Export page layout

pgiraud:
Will need more work later when when add more real content.


thinkhazard/static/less/common-variables.less
```diff
@@ -0,0 +1,6 @@
+@navbar-height: 60px;
+
+@level-high-color: #D91F2C;
+@level-medium-color: #D57C27;
+@level-low-color: #E0B025;
+@level-default-color: grey;
```
elemoine: It would be good to add the "newline" character at the end of this file.

thinkhazard/static/less/common.less
```diff
@@ -17,4 +17,8 @@ ul.horizontal li {
 // overrides for typeahead-bootstrap-less
 .form-group-lg .twitter-typeahead .tt-dropdown-menu {
   font-size: @font-size-large;
+}
+
+.navbar {
+  border: 0;
 }
```
elemoine: Again, no newline at end of file.

elemoine:
Two minor comments, otherwise looks good to me.

PR #22, opened 2015-04-24
Change admindiv to administrativedivision in the code

elemoine:
This is as suggested by @pgiraud. Let's use the same terms everywhere to avoid any confusion.


fvanderbiest:
okay

PR #23, opened 2015-04-24
passing the administrative division id to the report page

fvanderbiest:


thinkhazard/views.py
```diff
@@ -37,6 +37,15 @@ def index(request):
...
 @view_config(route_name='report', renderer='templates/report.jinja2')
 def report(request):
+    if 'administrativedivision_id' not in request.params:
+        raise HTTPBadRequest(
+            detail='parameter "administrativedivision_id" is missing')
+
+    id = request.params['administrativedivision_id']
+
+    place = DBSession.query(AdministrativeDivision).filter(
+        AdministrativeDivision.id == id).one()
```
pgiraud: What about using `DBSession.query(Table).get(id)`?

pgiraud:
I added a minor comment. Otherwise looks good to me.

PR #24, opened 2015-04-24
LESS optimizations

pgiraud:
The goal of this pull request is to get a smaller css file by removing any dependence on openlayers and by loading only the classes we use from bootstrap.

### DIFF
--- a/thinkhazard/static/less/bootstrap.less
+++ b/thinkhazard/static/less/bootstrap.less
@@ -1,0 +1,50 @@
+// Core variables and mixins
+@import "../../../node_modules/bootstrap/less/variables.less";
+@import "../../../node_modules/bootstrap/less/mixins.less";
+
+// Reset and dependencies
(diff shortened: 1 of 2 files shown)


ingenieroariel:
If we are not using open layers anymore, how are the maps being loaded?

pgiraud:
@ingenieroariel because we don't need any interaction on the map (no zoom or pan) we decided to get rid of OpenLayers. The map will be request to the service as a simple image.
This will eventually lighten the code and decrease the application complexity.

ingenieroariel:
Thanks for the explanation Pierre. It makes sense.

elemoine:
We still load ol in the page: https://github.com/GFDRR/thinkhazard/blob/master/thinkhazard/templates/report.jinja2#L101-L107. This should be removed as well.

elemoine:
The OpenLayers 3 code was removed with https://github.com/GFDRR/thinkhazard/pull/68. Please rebase this and merge. Thanks.

PR #25, opened 2015-04-24
Using a custom template for typeahead, makes search results more readable

pgiraud:
Changed order of administrative levels and make the higher level more obvious in the search results so that it's more intuitive for the user.
See screenshots.

Before:
screenshot from 2015-04-24 15 59 33

After:
screenshot from 2015-04-24 16 00 25

### DIFF
--- a/thinkhazard/static/js/search.js
+++ b/thinkhazard/static/js/search.js
@@ -23,20 +23,32 @@
 
   engine.initialize();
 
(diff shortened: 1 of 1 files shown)


ingenieroariel:
Excellent!

elemoine:
I merged this for @pgiraud, who is away for one week.

PR #27, opened 2015-04-30
Create map and UTFGrid using Mapnik

elemoine:
This PR adds a map to the report page. The map has two layers: a [NaturalEarth](http://www.naturalearthdata.com/) layer as the background layer and an "hazard" layer based on the administrative division and hazard data.

The map is created with Mapnik. UTFGrid is used to provide an "interactive" map, with the mouse cursor changing when over a subdivision. Subdivisions can be clicked to load a new hazard report.

Note that the "hazard" layer doesn't yet have the correct colors (based on the hazard category of each division). This PR just focuses on adding a map to the report page. And I'd like to address the rest with separate PRs.

Please review. In the mean time I'm going to test the branch on the demo server.

### DIFF
--- a/thinkhazard/templates/base.jinja2
+++ b/thinkhazard/templates/base.jinja2
@@ -29,7 +29,7 @@
       var app = {};
       app.administrativedivisionUrl = '{{'administrativedivision'|route_url}}';
       app.reportpageUrl = '{{'report'|route_url}}';
-      app.mapImgUrl = '{{'map'|route_url(format='png')}}';
+      app.mapImgUrl = '{{'map'|route_url(format='jpeg')}}';
       app.mapUtfUrl = '{{'map'|route_url(format='json')}}';
       {%- if division %}
       app.divisionCode = {{division.code}};
--- a/thinkhazard/templates/base.jinja2
+++ b/thinkhazard/templates/base.jinja2
@@ -29,6 +29,13 @@
       var app = {};
       app.administrativedivisionUrl = '{{'administrativedivision'|route_url}}';
       app.reportpageUrl = '{{'report'|route_url}}';
+      app.mapImgUrl = '{{'map'|route_url(format='png')}}';
+      app.mapUtfUrl = '{{'map'|route_url(format='json')}}';
+      {%- if division %}
+      app.divisionCode = {{division.code}};
+      {%- else %}
+      app.divisionCode = undefined;
+      {%- endif %}
     </script>
     <script src="{{('thinkhazard:static/js/%s.js' % self.name())|static_url}}"></script>
     <script src="{{'thinkhazard:static/js/search.js'|static_url}}"></script>
--- a/Makefile
+++ b/Makefile
@@ -95,7 +95,7 @@ thinkhazard/static/build/%.css: $(LESS_FILES) .build/node_modules.timestamp
 
 .build/venv:
 	mkdir -p $(dir $@)
-	virtualenv --no-site-packages .build/venv
+	virtualenv --system-site-packages .build/venv
 
 .build/venv/thinkhazard.wsgi: thinkhazard.wsgi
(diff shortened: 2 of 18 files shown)


elemoine:
Things seem to work as expected on the demo server. Example URLs: http://wb-thinkhazard.dev.sig.cloud.camptocamp.net/thinkhazard/wsgi/report?divisioncode=169 and http://wb-thinkhazard.dev.sig.cloud.camptocamp.net/thinkhazard/wsgi/report?divisioncode=122.

elemoine:
Things that still need work:

* Multipolygon divisions that span a lot of ground on the earth lead to maps that are very zoomed out. (See [France](http://wb-thinkhazard.dev.sig.cloud.camptocamp.net/thinkhazard/wsgi/report?divisioncode=85) for example).
* Correctly handle admin 2. Currently, they are not correctly displayed, because they don't have sub-divisions.
* Actually draw a classified map.
* Try to use `mapnik.PythonDatasource` rather than `MemoryDatasource` for better performance. I've experienced Mapnik crashes with `PythonDatasource` but that might be related to the data not having the correct SRID in the database.

Again, I'd like to address these things with separate PRs.

thinkhazard/templates/base.jinja2
```diff
@@ -29,6 +29,13 @@
       var app = {};
       app.administrativedivisionUrl = '{{'administrativedivision'|route_url}}';
       app.reportpageUrl = '{{'report'|route_url}}';
+      app.mapImgUrl = '{{'map'|route_url(format='png')}}';
```
fvanderbiest: did you try jpeg instead ? 

It might not look as good, but loading time could be improved.

elemoine: No I haven't. But I could.

fvanderbiest:
> Multipolygon divisions that span a lot of ground on the earth lead to maps that are very zoomed out. 

True. And the result might not be relevant at all.
eg: risks are clearly not the same in French Guiana and Metropolitan France

fvanderbiest:
Looks good otherwise, please merge.

elemoine:
The map service now supports png as well as jpeg, and jpeg is what the interface uses at this point.

elemoine:
Thanks for the review @fvanderbiest. Merging…

vdeparday:
Are we using Mapnik ? It is actually what we wanted to discuss during the last call instead of geoserver with @ingenieroariel  and we didn't have time to. It could be good option to generate nice looking maps.

ingenieroariel:
Yes Vivien, it is using Mapnik. I agree with you: we are very happy with
this choice.

elemoine:
We decided to give Mapnik a try for two (main) reasons: (a) UTFGrid, which is a good option for the map interactivity we need, and (b) be able to programmatically create maps based on the selected administrative division and hazard type. We do SQL queries using SQLAlchemy and draw the resulting features using Mapnik. This turns out to be a very good option. What we need to test though is whether Mapnik, and our use of MemoryDatasource or PythonDatasource, will work and scale well.

springmeyer:
I'll add a few thoughts:

  - There is a boost bug that impacts the Mapnik python plugin that I mention at https://github.com/mapnik/non-core-plugins/issues/1#issuecomment-104674614
  - Mapnik 3 will be released in a few weeks: http://mapnik.org/news/2015/05/21/release-3.0.0-rc2/
  - Mapnik 3 will be the first Mapnik release to provide binary wheels for easy python installations within virtualenv
  - The binary wheels will carry the key patch to prevent the boost python crash: https://github.com/mapbox/mason/blob/boost_libpython-1.57.0/patch.diff
  - The Feature and Geometry API changed in Mapnik 3.x but I'd be happy to help you upgrade if you have any trouble. See [this test](https://github.com/mapnik/python-mapnik/blob/master/test/python_tests/geometry_io_test.py) for examples

PR #72, opened 2015-05-11
JavaScript code refactoring

elemoine:
This PR makes some refactoring to the JavaScript code.

### DIFF
--- a/thinkhazard/static/js/report.js
+++ b/thinkhazard/static/js/report.js
@@ -1,12 +1,15 @@
 (function() {
 
+  // UTFGrid data
   var grid;
   var keys;
   var data;
-  var currentCursor;
-  var timeoutId = -1;
+
+  // Map element selector
   var mapSelector = '#map';
-  var hazardType;
+
+  // Array used as a temporary storage for event offsetX and offsetY
+  var offsets = new Array(2);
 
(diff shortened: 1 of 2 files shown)


thinkhazard/static/js/report.js
```diff
@@ -16,21 +19,27 @@
...
+  var timeoutId = -1;
   $(window).resize(function() {
     if (timeoutId != -1) {
       clearTimeout(timeoutId);
     }
-    timeoutId = setTimeout(addMap, 500);
+    timeoutId = setTimeout(function() {
+      addMap(hazardType);
+      timeoutId = -1;
+    }, 500);
```
pgiraud: Do we really need to affect a "-1" value to the timeoutId?
What if we call `clearTimeout` each time?

elemoine: I don't know what `clearTimeout` does when you pass it an invalid timeoutId value.

pgiraud: It fails silently:
```javascript
var foo;
clearTimeout(foo);
```

elemoine: Ok, so unconditionally calling `clearTimeout` should work.

elemoine: Changed as requested.

PR #78, opened 2015-05-11
Add a button to zoom out to upper-level area

elemoine:
This PR adds a button to the map to zoom out to the parent division.

@pgiraud, 9f3b70a adds `parent_division` to the context passed to the report page template. `parent_division` is an instance of `AdministrativeDivision`, so it has `code` and `name` properties.


pgiraud:
This pull request is now ready for review.

thinkhazard/templates/report.jinja2
```diff
@@ -36,6 +36,7 @@
           <a id="pdf-link" href><img src="{{('thinkhazard:static/images/download.png'|static_url)}}">Download PDF</a>
           <h3>Report on Hazards</h3>
           <h2>{{ division.name }}</h2>
+          <h4><a href="{{'report'|route_url}}?divisioncode={{ parent_division.code }}" title="Report on hazards in {{ parent_division.name }}">{{ parent_division.name}}</a>&nbsp;</h4>
```
elemoine: The official way to do that is:

```
{{ 'report'|route_url(_query={'divisioncode': parent_divisioncode}) }}
```

elemoine:
I just added a comment regarding the way to generate the "report" URLs.

thinkhazard/templates/report.jinja2
```diff
@@ -89,6 +90,12 @@
     </div>
     <div id="map-wrapper" class="">
       <div id="map">
+        {% if parent_division %}
+        <a id="map-drillup" class="btn btn-default" href="{{ 'report'|route_url(_query={'divisioncode': parent_division.code}) }}">
+          <img src="{{('thinkhazard:static/images/zoom-out.png'|static_url)}}">
```
elemoine: We should use `static_url` here.

elemoine:
LGTM. Please merge.

thinkhazard/views/report.py
```diff
@@ -64,4 +68,5 @@ def report(request):
     hazard_data = hazard_data.values()
     hazard_data = sorted(hazard_data, key=lambda d: d['categorytype'].order)
 
-    return {'hazards': hazard_data, 'division': division}
+    return {'hazards': hazard_data, 'division': division,
+            'parent_division': division.parent}
```
fvanderbiest: Do we really need to pass ```parent_division``` to the template ?

elemoine: It is used in `report.jinja2`.

fvanderbiest: True, but the template could probably also call division.parent instead ?

elemoine: You're right. I personally like this way because it justifies the use of a `join` in the SQL query.

Issue #79, opened 2015-05-11
Preserve selected hazard type when changing administrative division


elemoine:
Currently the hazard type selection is lost when clicking on the map to change to a sub-division. For example:

- Load http://wb-thinkhazard.dev.sig.cloud.camptocamp.net/thinkhazard/wsgi/report?divisioncode=169
- Select Earthquake
- Change to administrative division "Centre" by clicking to the corresponding sub-division on the map
- The "Overview page" instead of the "Earthquake page" is displayed!

PR #88, opened 2015-05-12
Add hash in url for selected hazard type

pgiraud:
Allows to keep hazard when drilling up/down to other division

Fixes https://github.com/GFDRR/thinkhazard/issues/79.

### DIFF
--- a/thinkhazard/static/js/report.js
+++ b/thinkhazard/static/js/report.js
@@ -21,10 +21,18 @@
   // Reload the map when a hazard type is selected
   var hazardType;
   $('#hazard-types-list a[data-toggle="tab"]').on('shown.bs.tab', function(e) {
-    var target = this.href.split('#');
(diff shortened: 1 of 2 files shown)


thinkhazard/static/js/report.js
```diff
@@ -18,14 +18,33 @@
...
   });
 
+  /**
+   * Take the hash from URL to activate the tab corresponding to the hazard
+   * type.
+   */
+  function showTab() {
+    var tab = window.location.hash.replace(prefix, '');
+    $('#hazard-types-list a[href=' + tab + ']').tab('show');
+  }
```
elemoine: I placed all the functions at the end of the file.

elemoine:
Please move the `showTab` function down with the other functions before merging. Otherwise looks good. This is a major usability improvement.

PR #94, opened 2015-05-12
Show legend only when hazard type is selected

pgiraud:


elemoine:

PR #98, opened 2015-05-12
Prevent scroll when tab changes

### LINKED ISSUE #97: Prevent scroll when switching from a tab to an other
This is due to our use of hash # in urls which correspond to anchors in the HTML.

screenshot from 2015-05-12 13 40 24

pgiraud:
Fixes #97.


elemoine:
Please merge.

Issue #87, opened 2015-05-12
Use a buffer for the map


elemoine:
Currently the map extent is that of the selected administrative division. For better maps a buffer should be added around the the extent of administrative division.

PR #99, opened 2015-05-12
Use a buffer for the map

elemoine:
A buffer is now added around the extent of the admin division. The size of the buffer is 1/8th the size of the admin division extent.

Fixes https://github.com/GFDRR/thinkhazard/issues/87.


pgiraud:
LGTM. Please merge.

Issue #41, opened 2015-05-04
Map : base map layer


fjacon:

elemoine:
To be discussed. We currently use the highest resolution map of NaturalEarth.

ingenieroariel:
I really like natural earth at the continent and country level. Can we use
for now something like Stamen Toner when we are below an specific zoom
level?

http://maps.stamen.com/toner/#11/10.9820/-74.8299

For the second iteration we can think better about designing a custom base
map.

-a

elemoine:
@fvanderbiest, @pgiraud and discussed the « background layer » issue yesterday. Obviously, we agree that the resolution of the NaturalEarth layer is too low. We've identified five possible options:

1. Find another (possibly commercial) shaded relief layer
2. Generate our own shaded relief layer (TIF) from some DEM data
3. Generate our own shaded relief layer (TIF) from tiles (MapBox tiles for example)
4. Use a vector-based layer with no relief (using NaturalEarth shapefiles maybe)
5. Use tiles, from a tile web service (MapBox, OSM, …)

After discussion, we think that 5. is the most straightforward path. It is also the most flexible, as we can use OSM now, and switch to another, better looking, layer at any time in the future. This would mean adding OpenLayers back but that is not an issue at all.

fvanderbiest:
Related to this issue, Nicolas says elsewhere:

```
Would it be possible to use tiles from MapQuest Open?

It seems that two free tilesets are available:
 * MapQuest-OpenStreetMap Tiles (mentioned by Stuart)
 * MapQuest Open Aerial Tiles (base map used in this html template)

Maybe we could use a different "look" by administrative division level, so that we are able to identify the level of administrative division just by looking at the map?

For example :
 * Level 0, MapQuest Open Aerial : http://mapq.st/1F6qDHC (or current base map, which seems to be Natural Earth)
 * Level 1, MapQuest OpenStreetMap : http://mapq.st/1KI7gGd
 * Level 2, MapQuest OpenStreetMap : http://mapq.st/1FjOiHP
```

ingenieroariel:
I will talk to the guys at Mapbox to see if we can use mapbox streets.

ingenieroariel:
Actually, now that I think about it ....

Since we have a fixed set of viewports, we can look into adapting OSM
bright for mapnik and just cache all the potential requests in ram or disk.
I believe at that zoom level the shaded relief can be optional (but we
should discuss it next week).

[...]

Issue #34, opened 2015-05-04
Basemap resolution is super low for the city level


vdeparday:
Searching for Ben Slimane, Morocco or Colombo, Sri Lanka, the background map/base layer is very low and show big pixels.

On a related topic, have we decided what the basemap would be?

elemoine:
We use NaturalEarth for the moment. See https://github.com/GFDRR/thinkhazard/blob/master/doc/data.md#raster.

pgiraud:
See #41.

PR #136, opened 2015-05-25
Map displayed using OL3 [WIP]

pgiraud:
Note: work is still in progress

With this pull request, the map is rendered using OpenLayers3 which allows us to use any tile service as base map.
See #34 and #41.
For now, the chosen basemap is `Stamen Watercolor` because it's not intrusive (no labels and pastel colors). It can be changed eventually.

The overlay layer showing the administrative divisions and hazard choropleths is the result of a request to the mapnik service (with almost no change). This image is computed so that it covers the administrative area but not more.

What's missing for now is the interaction (drill down).

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "jquery": "~2.1.3",
     "jshint": "~2.5.1",
     "less": "~1.7.5",
-    "openlayers": "~3.3.0",
+    "openlayers": "~3.6.0",
     "typeahead.js": "~0.10.5",
     "typeahead.js-bootstrap3.less": "hyspace/typeahead.js-bootstrap3.less#a512760"
   }
--- a/package.json
+++ b/package.json
@@ -22,6 +22,7 @@
     "jquery": "~2.1.3",
     "jshint": "~2.5.1",
     "less": "~1.7.5",
+    "openlayers": "~3.3.0",
     "typeahead.js": "~0.10.5",
     "typeahead.js-bootstrap3.less": "hyspace/typeahead.js-bootstrap3.less#a512760"
   }
--- a/thinkhazard/static/js/report.js
+++ b/thinkhazard/static/js/report.js
@@ -3,12 +3,6 @@
   // Map element selector
   var mapSelector = '#map';
 
-  // The hazard layer
-  var hazardLayer;
-
-  // Source for the hazard layer
-  var hazardLayerSource;
-
   // Array used as a temporary storage for event offsetX and offsetY
   var offsets = new Array(2);
 
@@ -27,6 +21,9 @@
     view: new ol.View({ })
   });
   map.getView().fitExtent(divisionBounds, map.getSize());
+
+  var hazardLayer = new ol.layer.Image({});
+  map.addLayer(hazardLayer);
 
   // Change the tab to active in the tablist when a item is selected
   // in the "overview" tabpanel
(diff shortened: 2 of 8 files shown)


package.json
```diff
@@ -22,6 +22,7 @@
     "jquery": "~2.1.3",
     "jshint": "~2.5.1",
     "less": "~1.7.5",
+    "openlayers": "~3.3.0",
```
elemoine: We can use 3.5.0 now.

thinkhazard/templates/report.jinja2
```diff
@@ -126,3 +125,14 @@
...
 {% endblock %}
+
+{% block scripts %}
+{% if debug %}
+    <script src="{{('%s/openlayers/dist/ol-debug.js' % node_modules)|static_url}}"></script>
+{% else %}
+    <script src="{{('%s/openlayers/dist/ol.js' % node_modules)|static_url}}"></script>
+{% endif %}
+    <script>
+      var division_bounds = {{bounds}};
```
elemoine: Please use the camelCase notation for js variables.

thinkhazard/views/report.py
```diff
@@ -68,5 +70,13 @@ def report(request):
     hazard_data = hazard_data.values()
     hazard_data = sorted(hazard_data, key=lambda d: d['categorytype'].order)
 
+    # Get the geometry for division and compute its extent
+    # FIXME use box2d
+    division_geometry = DBSession.query(AdministrativeDivision.geom) \
+        .filter(AdministrativeDivision.code == division_code).one()
```
elemoine: We already retrieved the administrative division with Query 1, no?

thinkhazard/templates/base.jinja2
```diff
@@ -31,7 +31,7 @@
       var app = {};
       app.administrativedivisionUrl = '{{'administrativedivision'|route_url}}';
       app.reportpageUrl = '{{'report'|route_url}}';
-      app.mapImgUrl = '{{'map'|route_url(format='jpeg')}}';
+      app.mapImgUrl = '{{'map'|route_url(format='png')}}';
```
elemoine: @fvanderbiest was the one suggesting to use jpeg. Why did you change it?

thinkhazard/static/js/report.js
```diff
@@ -210,7 +97,40 @@
...
+    showHazard();
     $('#legend').hide();
   }
+
+  /**
+   * Adds or replaces the hazard image layer.
+   */
+  function showHazard(hazardType) {
+    if (hazardLayer) {
+      map.removeLayer(hazardLayer);
```
elemoine: There's a `setSource` method that we could use instead of re-creating the layer.

ingenieroariel: Can we use this basemap instead of stame watercolor?

https://a.tiles.mapbox.com/v4/ingenieroariel.m9a2h374/page.html?access_token=pk.eyJ1IjoiaW5nZW5pZXJvYXJpZWwiLCJhIjoibXhDZ3pIMCJ9.qTmPYCbnUKtaNFkvKKysAQ#3/0.00/0.00

thinkhazard/templates/base.jinja2
pgiraud: This is because I want the overlay layer to be transparent.

pgiraud:
Obsolete pull request. See #143.

PR #141, opened 2015-06-11
Fix wrong url pattern matching regexp

pgiraud:


elemoine:
Line is too long. Otherwise looks good to me.

pgiraud:
Fixed. Merging.

Issue #145, opened 2015-06-18
Map drill down


pgiraud:
Add behavior to drill down when clicking on subdivision.
We can use `app.reportpageUrl`.

pgiraud:
Fixed with #198

PR #169, opened 2015-06-19
drill down on subdivision clicked

fvanderbiest:
Proposal for https://github.com/GFDRR/thinkhazard/issues/145

### DIFF
--- a/thinkhazard/static/js/report.js
+++ b/thinkhazard/static/js/report.js
@@ -22,7 +22,7 @@
   map.getView().fitExtent(app.divisionBounds, map.getSize());
 
   var vectorLayer = addVectorLayer(map, app.mapUrl);
-  addSelectInteraction(map, vectorLayer);
+  getSelectInteraction(map, vectorLayer);
 
   // change mouse cursor when over division
   map.on('pointermove', function(e) {
@@ -95,7 +95,7 @@
    * @param {ol.layer.Vector} layer
    * @return {ol.interaction.Select}
    */
-  function addSelectInteraction(map, layer) {
+  function getSelectInteractions(map, layer) {
     var textStyle = new ol.style.Text({
       text: '',
       scale: 1.2,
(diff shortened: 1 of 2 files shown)


thinkhazard/static/js/report.js
```diff
@@ -129,13 +129,31 @@
...
+    ];
+    clickInteraction.on('select', function(e) {
+      var feature = e.selected[0];
+      if (!feature) {
+        return;
+      }
+      var code = feature.getProperties().code;
+      if (code) {
+        window.location = app.reportpageUrl.replace(
+        "__divisioncode__", code);
```
elemoine: Please indent.

thinkhazard/static/js/report.js
```diff
@@ -129,13 +129,31 @@
...
+      }
+    });
+    map.getInteractions().extend(interactions);
+    return interactions;
```
elemoine: We should change the function name to `getSelectInteractions`.

thinkhazard/static/js/report.js
```diff
@@ -129,13 +129,31 @@
...
+        style: styleFn
+      }),
+      clickInteraction
+    ];
+    clickInteraction.on('select', function(e) {
+      var feature = e.selected[0];
+      if (!feature) {
+        return;
+      }
+      var code = feature.getProperties().code;
```
elemoine: Works by chance. Use `feature.get('code')`.

thinkhazard/static/js/report.js
```diff
@@ -129,13 +129,31 @@
...
+        layers: [layer],
+        condition: ol.events.condition.pointerMove,
+        style: styleFn
+      }),
+      clickInteraction
+    ];
+    clickInteraction.on('select', function(e) {
+      var feature = e.selected[0];
+      if (!feature) {
+        return;
```
elemoine: Can this happen?

fvanderbiest: I don't know, hence I'm cautious.

elemoine: I disagree with this way of thinking :) To me either you know that the `selected` array always has elements and you don't use a test, or you know that there are cases where the `selected` array is empty and you use a test. But you should not add a test because you don't know. I'd argue that this is an anti-pattern. 

And what about `e.selected`? You access its first element without testing that it's not `undefined`? Do you know that it cannot be `undefined`?

thinkhazard/static/js/report.js
elemoine: Sorry I meant `addSelectInteractions`. The plural was the important thing.

thinkhazard/static/js/report.js
```diff
@@ -22,7 +22,7 @@
   map.getView().fitExtent(app.divisionBounds, map.getSize());
 
   var vectorLayer = addVectorLayer(map, app.mapUrl);
-  addSelectInteraction(map, vectorLayer);
+  getSelectInteraction(map, vectorLayer);
```
elemoine: `s` missing at end of function name.

elemoine:
Please merge when the remaining comments are addressed. 

FWIW, I would not have used a `Select` interaction here. I'd just have used the following :


```js
map.on('click', function(e) {
  var feature = map.forEachFeatureAtPixel(e.pixel, function(feature) {
    return feature;
});
  if (feature) {
    //... 
 } 
});
```

pgiraud:
@fvanderbiest Do you think you can take 10 minutes to finish the work on this pull request?

fvanderbiest:
Sure

pgiraud:
Addressed in #198. Closing.

PR #174, opened 2015-06-22
Layout and style for hazard type icons

pgiraud:
We're still missing real icons, but this can be modified later.
Please review.


elemoine:

PR #179, opened 2015-06-23
Font family needs to be set only once

### LINKED ISSUE #178: Fonts families should be defined in one place
For example, we shouldn't need to define a font family for headers.
Font family for headers should be defined once and for all.

pgiraud:
Fixes #178


elemoine:
Looks much better.

PR #181, opened 2015-06-24
Rename watchless to watch

elemoine:
This renames the `watchless` target to `watch`, and makes it watch the entire `thinkhazard/static` dir instead of just the `thinkhazard/static/less` dir. It also runs `make build` instead of just rebuilding the CSS files.


elemoine:
@pgiraud, @marionb when this is merged the `watch` target should be used rather than `watchless`!

pgiraud:
Please merge.

PR #182, opened 2015-06-25
Style of Download and Overview buttons

marionb:


pgiraud:
In general, we'd better avoid using ids.
In the current case, the buttons are supposed to look exactly the same. We should use a shared class for both.

thinkhazard/static/less/report.less
```diff
@@ -229,3 +229,38 @@ body {
...
+
+.style-pdf-overview-btn {
+  padding: 7px 10px;
+  margin-top: 15px;
+  border: 1px solid #d9d9d9;
+  font-size: 14px;
+
+  -webkit-border-radius: 5px;
+     -moz-border-radius: 5px;
+          border-radius: 5px;
```
pgiraud: If I remember well, there's a mixin in bootstrap or less to do this more automatically.

thinkhazard/static/less/report.less
```diff
@@ -229,3 +229,38 @@ body {
...
+    color: @color;
+  }
+  &:active{
+    background-color:@color;
+    border-color: @color;
+    a {
+      color: #fff;
+    }
+
+   }
```
pgiraud: Indentation is weird here.

pgiraud:
Even more, I think we should override the "btn-xs" bootstrap class.

marionb:
In the document the overview button has an additional active property that is not described in the pdf-button. After overwriting the "btn-xs" can I additionally add this property to the button?

pgiraud:
Yes, we can add the active specific styling for this button.
But, I don't think active means ":active" here. It's more like telling the user that the report page is currently showing the overview or not. In my opinion, it would be better to remove the "overview" button when displaying the overview.
Let's forget this "active" thing for now. We'll discuss this with the designer.

thinkhazard/static/less/report.less
```diff
@@ -229,3 +229,25 @@ body {
     margin-bottom: 4px;
   }
 }
+
+.style-pdf-overview-btn {
+  padding: 7px 10px;
+  margin-top: 15px;
+  border: 1px solid #d9d9d9;
+  font-size: 14px;
```
pgiraud: This should be set globaly.

thinkhazard/static/less/report.less
```diff
@@ -229,3 +229,25 @@ body {
...
+  }
+}
+
+.btn-xs {
+  .style-pdf-overview-btn;
+}
+
+.btn-default {
+  .style-pdf-overview-btn;
+}
```
pgiraud: Are you sure you want to change .btn-default as well?
.btn-default is not meant to be used for sizes, only for colors

thinkhazard/static/less/report.less
```diff
@@ -229,3 +229,25 @@ body {
     margin-bottom: 4px;
   }
 }
+
+.style-pdf-overview-btn {
+  padding: 7px 10px;
+  margin-top: 15px;
+  border: 1px solid #d9d9d9;
```
pgiraud: The border should be set globaly.

thinkhazard/static/less/report.less
```diff
@@ -229,3 +229,25 @@ body {
     margin-bottom: 4px;
   }
 }
+
+.style-pdf-overview-btn {
+  padding: 7px 10px;
+  margin-top: 15px;
```
pgiraud: The margin-top shouldn't not be set in the button itself. It actually means that the download button has to be 15px away from the navbar. But I wouldn't care much for now.

thinkhazard/static/less/report.less
```diff
@@ -229,3 +229,25 @@ body {
...
+  border: 1px solid #d9d9d9;
+  font-size: 14px;
+  .border-top-radius(5px);
+  .border-right-radius(5px);
+  .border-left-radius(5px);
+  .border-bottom-radius(5px);
```
pgiraud: I don't think those mixins are useful for what you want to do.
The following is enough in my opinion:
```css
border-radius: 5px;
```

thinkhazard/templates/report.jinja2
```diff
@@ -62,10 +62,10 @@ Think Hazard - {{ division.name}}
 
       <!-- Division name -->
       <h2>
-        <a id="division-name" href="{{ overview_url }}" title="Go back to overview page">{{ division.name }}</a>
+        <a href="{{ overview_url }}" title="Go back to overview page">{{ division.name }}</a>
       </h2>
 
-      <a class="btn btn-default btn-xs" href="{{ overview_url }}" title="Go back to overview page">Overview</a>
+      <a id="overview-link" class="btn btn-default btn-xs" href="{{ overview_url }}" title="Go back to overview page"><i class="fa fa-home"></i>Overview</a>
```
pgiraud: Please add a space between icon and "overview" text.

thinkhazard/static/less/report.less
```diff
@@ -250,3 +250,11 @@ body {
 .further-resources-level(MED);
 .further-resources-level(LOW);
 .further-resources-level(NPR);
+
+#pdf-link, #overview-link {
+  border: 1px solid #d9d9d9;
+  &:hover {
+    .box-shadow(5px 5px 5px rgba(0, 0, 0, 0.15));
```
pgiraud: In my opinion, the 5px dropshadow should be:
```less
.box-shadow(0 0 5px 0 rgba(0, 0, 0, 0.15));
```

thinkhazard/static/less/common.less
```diff
@@ -17,6 +17,7 @@
 @input-border-focus: #58595b;
 @input-border-radius: 5px;
 
+@border-radius-small: @input-border-radius;
```
pgiraud: I wouldn't reuse the `@input-border-radius` here. This variable is used in Bootstrap. This means that any input will have a radius.
We want to do the same thing for the small buttons. So actually, what we want to override here is: `@btn-border-radius-small`.
So I would write the following instead:
```less
@btn-border-radius-small: 5px;
```
I know, I'm picky.

pgiraud:
Ok, it looks good. Please merge. Thanks.

PR #184, opened 2015-06-25
Changing the background and list styling of further resources

marionb:


thinkhazard/static/less/report.less
```diff
@@ -223,9 +222,45 @@ body {
...
+    }
+  }
+}
+
+.further-resources-level(@level) {
+  @color: "level-@{level}-color";
+  .level-@{level} {
+    #further-resources li {
+      color: @@color;
+    }
```
pgiraud: You may want to apply this color only to ".fa-li".

thinkhazard/static/less/report.less
```diff
@@ -223,9 +222,45 @@ body {
...
+    li {
+      margin-top: 15px;
+      margin-left: 10px;
+    }
+    a {
+      color: #035e8b;
+    }
+    a:hover {
+      text-decoration: underline;
+    }
```
pgiraud: _color_ and _text-decoration_ for links are to be set the whole application IMO.

pgiraud:
Are you able to make it so the arrows are aligned on the left (ie. below the "F")?

marionb:
I can Align the text with the "F" (by removing the margin from the ul) but I have not relay the arrows. Is there a trick for this?

pgiraud:
Can you show me?

pgiraud:
OK. So let's merge this one and address the alignment later.

PR #192, opened 2015-07-02
Add nav. bar links and add style

marionb:
Please review.

### DIFF
--- a/thinkhazard/static/less/common.less
+++ b/thinkhazard/static/less/common.less
@@ -39,6 +39,12 @@ ul.horizontal li {
   .box-shadow(0px 2px 10px -1px rgba(0, 0, 0, 0.33));
 }
 
+.divider-vertical {
+  border-left: 1px solid #d9d9d9;
+  margin-top: @navbar-padding-vertical;
+  margin-bottom: @navbar-padding-vertical;
+}
+
 .navbar-brand {
   font-family: 'variableblack', sans-serif;
   font-size: 25.2px;
--- a/thinkhazard/templates/report.jinja2
+++ b/thinkhazard/templates/report.jinja2
@@ -1,5 +1,7 @@
 {% extends "base.jinja2" %}
 
+{% from 'navbar.jinja2' import navbar%}
+
 {% block name %}report{% endblock %}
 
 {% block title %}
@@ -26,14 +28,7 @@ Think Hazard - {{ division.name}}
           {% include 'search.jinja2' %}
         </div>
       </form>
-      <ul class="nav navbar-nav navbar-right">
-        <li class="disabled">
-          <a href="">About</a>
-        </li>
-        <li class="disabled">
-          <a href="">Contact us</a>
-        </li>
-      </ul>
+      {{ navbar() }}
     </div>
   </nav>
 
--- a/thinkhazard/static/less/bootstrap.less
+++ b/thinkhazard/static/less/bootstrap.less
@@ -18,7 +18,7 @@
 
 // Components
 /*@import "../../../node_modules/bootstrap/less/component-animations.less";*/
-/*@import "../../../node_modules/bootstrap/less/dropdowns.less";*/
+@import "../../../node_modules/bootstrap/less/dropdowns.less";
 @import "../../../node_modules/bootstrap/less/button-groups.less";
 /*@import "../../../node_modules/bootstrap/less/input-groups.less";*/
 @import "../../../node_modules/bootstrap/less/navs.less";
--- a/thinkhazard/static/less/report.less
+++ b/thinkhazard/static/less/report.less
@@ -6,8 +6,6 @@
 @level-NPR-color: #87988c;
 @level-no-data-color: #F2F2F2;
 
-@text-color: #333;
-
 body {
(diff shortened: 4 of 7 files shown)


thinkhazard/static/less/common.less
```diff
@@ -37,6 +37,13 @@ ul.horizontal li {
 .navbar {
   border: 0;
   .box-shadow(0px 2px 10px -1px rgba(0, 0, 0, 0.33));
+  li:last-child a{
+      padding-top: 0px;
+      padding-bottom: 0px;
+      margin-top: 29px;
+      margin-bottom: 29px;
+      border-left: 1px solid #d9d9d9;
```
pgiraud: It a bit ugly when on click on the button.
screenshot from 2015-07-02 15 05 25


It would be better to have a separate element for the divider than adding a border.

See https://www.google.fr/search?q=bootstrap+navbar+separator&oq=bootstrap+navbar+separa&aqs=chrome.1.69i57j69i59.9931j0j4&sourceid=chrome&es_sm=93&ie=UTF-8#q=bootstrap+navbar+divider&tbs=qdr:y
http://solvedstack.com/questions/vertical-divider-don-t-work-in-bootstrap-3

thinkhazard/static/less/common.less
```diff
@@ -37,6 +37,13 @@ ul.horizontal li {
 .navbar {
   border: 0;
   .box-shadow(0px 2px 10px -1px rgba(0, 0, 0, 0.33));
+  li:last-child a{
+      padding-top: 0px;
+      padding-bottom: 0px;
+      margin-top: 29px;
+      margin-bottom: 29px;
```
pgiraud: @marionb What's the purpose of changing padding and margin here?

thinkhazard/templates/report.jinja2
```diff
@@ -31,8 +31,20 @@ Think Hazard - {{ division.name}}
...
+        </li>
+        <li class="disabled">
           <a href="">Contact us</a>
         </li>
+        <li class="dropdown disabled">
+          <a href="" data-toggle="dropdown" class="dropdown-toggle">English <i class="fa fa-chevron-down"></i></a>
+          <ul class="dropdown-menu">
+            <li><a href="">English</a></li>
+          </ul>
+        </li>
```
pgiraud: Since the link in the navbar are supposed to be the same in index and report pages, we should put this in a separate (included) template.

thinkhazard/static/less/common.less
marionb: Do you mean adding the separator by adding a divider-vertical class instead of accessing the last element? This is what is don in the link you provided, but in the end they also add a border In the links you they also add it as a border.

thinkhazard/static/less/common.less
marionb: I changed the padding because I added the separator as a border. But the border has the same height as the text + the padding which makes the border too high (compared to what is on the mokup).
separator
If the padding is 0 and the margin to the height of the padding this fixes the problem. Perhaps there is a better way to do this?

thinkhazard/static/less/common.less
pgiraud: Yes, the idea is to have a separate element to show the divider. The divider and the language button should be different elements no matter how each is styled.
Using borders to style the divider isn't a problem.

thinkhazard/static/less/common.less
pgiraud: If you use a separate element for the divider, you should be able to remove the paddings and margins here.

marionb:
Please review, and let me know if adding the navbar items using a macro is OK or if there is a better way to do this.

pgiraud:
Is there any reason why you didn't use `block` and `super()` as I suggested? Have you had troubles using this technique?

thinkhazard/static/less/common.less
```diff
@@ -35,10 +35,16 @@ ul.horizontal li {
 }
 
 .navbar {
+  display: block;
```
pgiraud: Is this required?

thinkhazard/static/less/common.less
```diff
@@ -35,10 +35,16 @@ ul.horizontal li {
...
   border: 0;
   .box-shadow(0px 2px 10px -1px rgba(0, 0, 0, 0.33));
 }
 
+.divider-vertical {
+  border-left: 1px solid #d9d9d9;
+  margin: 29px 0px;
```
pgiraud: I'd prefer the vertical padding to be computed so that we can change the `@navbar-height` anytime.

If you take the `.navbar>li>a` as an example, the vertical padding take `@navbar-padding-vertical`.
https://github.com/twbs/bootstrap/blob/master/less/navbar.less#L273

This is transform to `29.5px` after less build.

thinkhazard/static/less/common.less
marionb: It says in the Document under Menu items display: block. So I understood that it is required.

pgiraud: Not required. Please remove it.

pgiraud:
Thanks. Please merge.

PR #198, opened 2015-07-06
Adding support for drill-down on map

pgiraud:
Replaces #169.
Please review.

### DIFF
--- a/thinkhazard/static/js/report.js
+++ b/thinkhazard/static/js/report.js
@@ -31,6 +31,17 @@
     map.getTargetElement().style.cursor = hit ? 'zoom-in' : '';
   });
 
+  // drill down
+  map.on('click', function(e) {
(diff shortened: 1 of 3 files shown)


thinkhazard/static/js/report.js
```diff
@@ -31,6 +31,19 @@
...
   });
 
+  // drill down
+  map.on('click', function(e) {
+    var feature = map.forEachFeatureAtPixel(e.pixel, function(feature) {
+      return feature;
+    });
+    if (feature) {
+      var code = feature.get('code');
+      if (code) {
```
elemoine: Out of curiosity, are there cases where a feature has no code?

elemoine:
I just added a minor comment. Please merge.

thinkhazard/static/js/report.js
pgiraud: I'm not completely sure what the server sends. But you're right, we can assume that this is always set. To be removed.

PR #199, opened 2015-07-06
Add style to hazard type and header

marionb:
Pleas review

### DIFF
--- a/thinkhazard/static/less/report.less
+++ b/thinkhazard/static/less/report.less
@@ -102,6 +102,23 @@ body {
   }
 }
 
+.hazard-title-level(@level) {
+  @color: "level-@{level}-color";
+  .level-@{level} {
+    .page-header {
+      border-color: @@color;
+      .level {
+        color: @@color;
+      }
+    }
+  }
+}
+
+.hazard-title-level(HIG);
+.hazard-title-level(MED);
+.hazard-title-level(LOW);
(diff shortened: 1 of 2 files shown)


thinkhazard/templates/report.jinja2
```diff
@@ -80,13 +80,13 @@ Think Hazard - {{ division.name}}
     <div class="row">
       <div class="col-md-7">
         {% if current_hazard %}
-          <h1 class="page-header detail level-{{ current_hazard.categorytype.mnemonic }}">
+          <h2 class="page-header detail level-{{ current_hazard.categorytype.mnemonic }}">
```
pgiraud: We should remove the ".level-XXX" class here since it's already defined at the container level.

thinkhazard/static/less/report.less
```diff
@@ -102,6 +102,23 @@ body {
...
 
+.hazard-title-level(@level) {
+  @color: "level-@{level}-color";
+  .level-@{level} {
+    .page-header {
+      border-color: @@color;
+    }
+    .level {
+      color: @@color;
+    }
```
pgiraud: This has an unexpected effect: the legend title is colored now. It shouldn't be.

pgiraud:
OK. You can merge now.
I'll fix misalignment later.
Also, I'll do some factorization of the less code.

PR #200, opened 2015-07-06
Pgiraud avoid drill down

### LINKED ISSUE #197: Don't pretend user can drill down on level 2
We should not show a zoom-in cursor, nor should we do on hover or click when we display a map for the higher subdivision level.

pgiraud:
Fixes #197.
#198 should be merged first.


pgiraud:
Pull request rebased. Please review.

PR #206, opened 2015-07-09
Fix map caption

### LINKED ISSUE #203: Mnemonic instead of complete name in map caption
In the map caption, only the name of the hazard type should be shown. Currently the mnemonic is displayed.
screenshot from 2015-07-09 10 05 15

pgiraud:
Fixes #203.


pgiraud:
Trivial. Merging.

PR #238, opened 2015-07-21
Icons

pgiraud:
Get rid of font-awesome and use the app web font for the icons.
Please review.


thinkhazard/static/less/icons.less
```diff
@@ -0,0 +1,81 @@
...
+  src: url('../webfonts/thinkhazard/fonts/thinkhazard.eot?-xr90r9');
+  src: url('../webfonts/thinkhazard/fonts/thinkhazard.eot?#iefix-xr90r9') format('embedded-opentype'),
+       url('../webfonts/thinkhazard/fonts/thinkhazard.ttf?-xr90r9') format('truetype'),
+       url('../webfonts/thinkhazard/fonts/thinkhazard.woff?-xr90r9') format('woff'),
+       url('../webfonts/thinkhazard/fonts/thinkhazard.svg?-xr90r9#thinkhazard') format('svg');
+  font-weight: normal;
+  font-style: normal;
+}
+
+[class^="icon-"], [class*=" icon-"] {
```
marionb: For better understanding: how do I understand this line?

marionb:
This looks good, I just added one question where I did not understand your code.
please merge.

thinkhazard/static/less/icons.less
pgiraud:
```
[class^="foo"]
```
The above means that any element with a class prefixed with `foo` will match.

```
[class*=" foo"]
```
The above means that any element with a class containing ` foo` (note the space) will match.

https://developer.mozilla.org/en-US/docs/Web/CSS/Attribute_selectors

PR #240, opened 2015-07-22
Add logo to footer and let it use all width

marionb:
Is it OK to use "top: 0.5em" (in report.less)?
Please review.

### DIFF
--- a/thinkhazard/static/less/report.less
+++ b/thinkhazard/static/less/report.less
@@ -223,6 +223,21 @@ body {
   }
 }
 
+footer {
+  margin-top: 20px;
+  display: table;
+  white-space: nowrap;
+  &:before, &:after {
+    border-top: 5px solid #e6e6e6;
+    content: '';
+    display: table-cell;
+    position: relative;
+    @footer-line-height: 5px;
+    top: calc(0.5em + @footer-line-height);
+    width: 50%;
+  }
+}
+
 .set-level-color(@level) {
   @color: "level-@{level}-color";
 
--- a/thinkhazard/static/less/report.less
+++ b/thinkhazard/static/less/report.less
@@ -136,8 +136,8 @@ body {
 
 .logo-gfdrr {
   display: block;
-  margin-left: auto;
-  margin-right: auto;
+  margin-left: 15px;
+  margin-right: 15px;
   width: 135px;
   height: 32px;
   background: url(../images/logo-gfdrr-small.png) 0 0 no-repeat;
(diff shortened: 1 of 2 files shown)


thinkhazard/static/less/report.less
```diff
@@ -252,6 +252,21 @@ body {
...
+  &:before, &:after {
+    border-top: 5px solid #e6e6e6;
+    content: '';
+    display: table-cell;
+    position: relative;
+    top:0.5em ;
+    width: 45%;
+  }
+  &:before { right: 1.5%; }
+  &:after { left: 1.5%; }
```
pgiraud: Those two lines should be removed.
If you look at the mockup, the line is aligned with the content above.

thinkhazard/static/less/report.less
```diff
@@ -252,6 +252,21 @@ body {
...
 
+footer {
+  display: table;
+  white-space: nowrap;
+  &:before, &:after {
+    border-top: 5px solid #e6e6e6;
+    content: '';
+    display: table-cell;
+    position: relative;
+    top:0.5em ;
```
pgiraud: Please leave space after colon. And remove space before semi-colon.

pgiraud: Here, I would use `calc`.
```
@footer-line-height: 5px;
top: calc(0.5em + @footer-line-height);
```

pgiraud:
I would be nice to have more space between main content and footer.
Adding `margin-top: 20px` to footer should do the job.

thinkhazard/static/less/report.less
marionb: If I use this instead of 'top: 0.5em;' the whole bar is moved too far down (see newly committed code) Do I have to add something else as well?

pgiraud: This looks good to me in my browsers (Chrome and Firefox).

screenshot from 2015-07-23 13 25 38

pgiraud: Forget my last comment. I know see the problem.

pgiraud: The problem is that calc doesn't work the way you expect in this case.
I just push a new commit to fix the problem.

pgiraud:
Can you please now try to find a workaround to re-add spacing around the logo? If this is too tricky, I can leave with the `right` and `left` values for `:before` and `:after`.

marionb:
Please review.

pgiraud:
Looks good.
The workaround works well.
Please merge.

PR #241, opened 2015-07-22
change color and syle of the thinkhazard logo

### LINKED ISSUE #230: Front page - Thinkhazard logo

marionb:
please review.
Fixes #230

### DIFF
--- a/thinkhazard/static/less/index.less
+++ b/thinkhazard/static/less/index.less
@@ -2,6 +2,7 @@
 
 .navbar-default {
   background-color: transparent;
+  margin-bottom: 27px;
 }
 
 // height of the search pane (ie. height of the background image)
@@ -46,6 +47,10 @@
 
 .hazard-types-list li .hazard-icon {
   .box-shadow(0px 0px 20px -1px rgba(0, 0, 0, 0.15));
+}
+
+.logo {
+  font-size: 48px;
 }
 
 .baseline {
--- a/thinkhazard/static/less/report.less
+++ b/thinkhazard/static/less/report.less
@@ -238,6 +238,11 @@ footer {
   }
 }
 
+.logo {  
+  font-size: 25px;
+  margin-right: 10px;
+}
+
 .set-level-color(@level) {
   @color: "level-@{level}-color";
 
--- a/thinkhazard/templates/common.jinja2
+++ b/thinkhazard/templates/common.jinja2
@@ -21,3 +21,9 @@
         </li>
       </ul>
 {% endmacro %}
+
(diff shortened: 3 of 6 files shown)


thinkhazard/static/less/index.less
```diff
@@ -47,3 +47,8 @@
 .hazard-types-list li .hazard-icon {
   .box-shadow(0px 0px 20px -1px rgba(0, 0, 0, 0.15));
 }
+
+.index-brand {
```
pgiraud: You don't need to create a new class here.
The `index.less` file is supposed to be compiled in `index.css` which is loaded only in `index.jinja2`.
It means that you should be able to override the font-size using `.logo-brand`.

thinkhazard/static/less/index.less
```diff
@@ -47,3 +47,8 @@
...
+  margin-top: 24px;
```
pgiraud: I'd prefer the margin-top to be set as a margin-bottom on the navbar.

thinkhazard/templates/index.jinja2
```diff
@@ -14,7 +14,9 @@
   </nav>
   <div class="container">
     <header class="text-center">
-      <img src="{{('thinkhazard:static/images/logo-big.png'|static_url)}}" title="Think hazard - Overcome risk" alt="Think hazard - Overcome risk" />
+      <a class="index-brand logo-brand" href="{{'index'|route_url}}">
+        <span class="logo-think">Think</span><span class="logo-hazard">Hazard</span><span class="logo-think">!</span>
```
pgiraud: Since we're reusing the same piece of code in both index and report, I'd suggest we factorize it.

thinkhazard/templates/common.jinja2
```diff
@@ -21,3 +21,9 @@
         </li>
       </ul>
 {% endmacro %}
+
+{% macro thinkhazardLogo() %}
```
pgiraud: It would be better to use snake case here (ie. thinkhazard_logo). Jinja2 templates can be considered as python code and by convention python variable and function names are written in snake case.

marionb:
Please review.

thinkhazard/static/less/report.less
```diff
@@ -238,6 +238,11 @@ footer {
   }
 }
 
+.logo {  
+  font-size: 25.2px;
```
pgiraud: Even if written in the document, this value looks weird to me. Please remove it.

pgiraud:
I just added a small comment. Please merge when addressed.

PR #244, opened 2015-07-27
Change hazard selection coloring

marionb:
please review.


pgiraud:
We probably don't want a bg color change on hover when a hazard is selected.

marionb:
Yes good point, I changed it.
Please review.

pgiraud:
Please merge.

PR #245, opened 2015-07-29
style hazard name

marionb:
Please review.

### DIFF
--- a/thinkhazard/static/less/common.less
+++ b/thinkhazard/static/less/common.less
@@ -6,6 +6,7 @@
 
 @font-size-base: 18px;
 @font-size-small: 14px;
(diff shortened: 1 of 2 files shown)


thinkhazard/static/less/common.less
```diff
@@ -67,6 +68,8 @@ ul.horizontal li {
 
 .page-header {
   position: relative;
+  color: #333;
+  font-weight: bold;
```
pgiraud: Instead of setting the font-weight here, I'd suggest we use `variablebold` font for headings.
Currently we use `variableregular` for h1 and h2 which is wrong in my opinion.

marionb:
Please  review.

pgiraud:
Please merge.

Issue #227, opened 2015-07-21
Report overview - Hazard level


pgiraud:
Item 4.

pgiraud:
Fixed via #246.

Issue #228, opened 2015-07-21
Report overview - Hazard section disabled


pgiraud:
Item 5.

pgiraud:
Fixed via #246.

PR #246, opened 2015-07-29
style hazard level

marionb:
Please review.

### DIFF
--- a/thinkhazard/static/less/report.less
+++ b/thinkhazard/static/less/report.less
@@ -249,10 +249,6 @@ footer {
   margin-right: 10px;
 }
 
-.level-no-data .page-header {
-  border-color: @level-overview-no-data-color;
-}
-
 .set-level-color(@level) {
   @color: "level-@{level}-color";
   @color-overview: "level-overview-@{level}-color";
@@ -300,3 +296,17 @@ footer {
 .set-level-color(MED);
 .set-level-color(LOW);
 .set-level-color(NPR);
+
+.page-header .level {
+  font-weight: 600;
+}
+
+.level-no-data .page-header {
+  border-color: @level-overview-no-data-color;
+  color: #ccc;
+  .level {
+    font-weight: 400;
+    color: #ccc;
+    font-size: @font-size-small;
+  }
+}
--- a/thinkhazard/static/less/common.less
+++ b/thinkhazard/static/less/common.less
@@ -70,6 +70,7 @@ ul.horizontal li {
     right: 0;
     bottom: 0;
     font-style: italic;
+    font-family: @font-family-base;
     padding-bottom: ((@line-height-computed / 2) - 1);
   }
 }


thinkhazard/static/less/report.less
```diff
@@ -249,6 +249,10 @@ footer {
   margin-right: 10px;
 }
 
+.hazard-active {
+    font-family: "Source Sans Pro Semibold";
```
pgiraud: This won't work unless this font is installed in the operating system of the user.
We load the font from `googleapis.com` so we only have the "Source Sans Pro" font-family available. See http://fonts.googleapis.com/css?family=Source+Sans+Pro

So if we want a bolder font-weight, let's use:
```
font-weight: bold;
```

thinkhazard/templates/report.jinja2
```diff
@@ -79,7 +79,7 @@ Think Hazard - {{ division.name}}
           <h2 class="page-header detail">
             {{ current_hazard.hazardtype.title }}
             <small>
-              Hazard level: <span class="level">{{ current_hazard.categorytype.title }}</span>
+              Hazard level: <span class="level hazard-active">{{ current_hazard.categorytype.title }}</span>
```
pgiraud: Is is really useful to add a new `hazard-active` class?
What about relying on the already existing level class?

thinkhazard/static/less/report.less
marionb: Is it OK to us bold even if semibold is given in the description?

thinkhazard/templates/report.jinja2
marionb: I added this class thinking it might be help full after when I change Item 5 the disabled hazard section, they also have the label class but a different font-style. I will see if I can do without it.

thinkhazard/static/less/report.less
pgiraud: https://www.google.com/fonts/specimen/Source+Sans+Pro
OK, so you can use `font-weight: 600`

thinkhazard/templates/report.jinja2
pgiraud: I think there's already a `level-no-data` in an ancestor which can help.

marionb:
Then I think we can set it in the mixin so we do not have to overwrite the property for the disabled section in item 5.
Please review.

pgiraud:
Can you please address #247 in the current pull request as well? I'd like to see the final result in a single pull request.

thinkhazard/static/less/report.less
```diff
@@ -274,18 +274,23 @@ footer {
...
+    .page-header {
+      &.overview {
+        border-color: #ccc;
+        &:hover {
+          border-color: fade(@@color, 50%);
+          background-color: @@color-overview;
+        }
+      }
+      .level {
+        font-weight: 600;
```
pgiraud: I'm not in favor of putting properties which are not specific to a level (no color change) in the mixin.

marionb:
#247 has just been merged. Do you mean #228?

pgiraud:
Yes, sorry. #228 is the issue that can be addressed in the current pull request.

pgiraud:
Looks good please merge.

marionb:
fixes #227 and #228

PR #248, opened 2015-07-30
Front page search and autocompletion

### LINKED ISSUE #233: Front page - Search bar + autocompletion

pgiraud:
Fixes #233.

### DIFF
--- a/thinkhazard/static/less/common.less
+++ b/thinkhazard/static/less/common.less
@@ -62,10 +62,6 @@ ul.horizontal li {
   padding-left: 0;
 }
 
-.form-control {
-  font-family: 'variablelight', sans-serif;
-}
-
 .page-header {
   position: relative;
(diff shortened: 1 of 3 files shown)


pgiraud:
Please review.

thinkhazard/static/less/common.less
```diff
@@ -194,12 +192,11 @@ ul.horizontal li {
 .tt-suggestion:hover {
   cursor: pointer;
   color: #fff;
```
marionb: If the background colour is white when hovering the text is almost not visible. Looking at the mock-up the text colour should stay the same as when not hovering.

pgiraud: Good catch.
I only tested selection with the keyboard. My bad.

marionb:
looks good, pleas merge.

PR #250, opened 2015-07-30
Show division outline on map

### LINKED ISSUE #229: Report overview - Administrative division polygon
Item 6.

pgiraud:
Fixes #229.
@elemoine it would be nice if you could have a review when you're back.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "jquery": "~2.1.3",
     "jshint": "~2.5.1",
     "less": "~1.7.5",
-    "openlayers": "~3.6.0",
+    "openlayers": "~3.8.2",
     "typeahead.js": "~0.11.1"
   }
 }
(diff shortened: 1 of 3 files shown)


pgiraud:
We unfortunately have an issue with OL3.
https://github.com/openlayers/ol3/issues/3941

elemoine:
> We unfortunately have an issue with OL3.
> openlayers/ol3#3941

The issue will be fixed in 3.8.0 (https://github.com/openlayers/ol3/pull/3883). So I guess we can wait a bit and use v3.8.0 instead of v3.7.0.

pgiraud:
According to the discussion we just had with @elemoine, this should be simplified. There's no real advantage to use 2 different vector layers.

pgiraud:
Please review again.

pgiraud:
I prefer to merge without any new review. Please reopen if you see issues.

PR #279, opened 2015-09-10
Sort hazards by hazard type instead of category type

### LINKED ISSUE #278: Consistent order of icons on all pages
Please can you make the order of icons consistent on all pages, to help user find the icons as easily as possible?
Based on relative data availability and frequency that people consider each hazard, the order should be as follows, from left to right on the home page and on recommendations page header

This should also be the order for the vertical list on recommendations page, unless the vertical stacking of hazards is determined by whether data is available (i.e., hazards with available data appear at top), in which case please use the following order as a secondary order, after ordering by data availability.

Flood
Earthquake
Drought
Cyclone 
Storm Surge
Tsunami 
Volcanic Ash
Landslide

Many Thanks

pgiraud:
Fixes #278.
Database needs an update though because order is based on hazard type table order which is not what Stuart expects.

### DIFF
--- a/thinkhazard/views/report.py
+++ b/thinkhazard/views/report.py
@@ -43,29 +43,27 @@ def report(request):
     # Get all the hazard types.
     hazardtype_query = DBSession.query(HazardType)
 
-    # Create a dict containing the data sent to the report template. The keys
-    # are the hazard type mnemonics.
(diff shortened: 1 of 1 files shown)


pgiraud:
@elemoine Would you please be kind to review this pull request?

pgiraud:
I merged it. If you have comments anyway please don't hesitate.

PR #307, opened 2015-11-26
Change search place holder to sthing more explicit

### LINKED ISSUE #306: Update of search bar - required in alpha 2 version
Please could we update the grey default text in the search bar from 'Ben Slimane' a more-commonly known location?
This could read 'Enter location... (e.g. Turkey, or Istanbul)'. 

We require this to be done in the alpha2 version as soon as possible, 

We have to share this with management this week to prepare for an internal review - management have fed back that the Ben Slimane example is somewhat confusing as they don't know immediately that it is the name of a location.

Many thanks
Stuart

pgiraud:
Fixes #306


pgiraud:
Trivial. Merging.

PR #311, opened 2015-12-04
Adding a map scale

### LINKED ISSUE #301: Map scale
Further to remote user testing, it is required that a map scale be included on the map. I'll add this feature in the form of a graphic scale located in the bottom left-hand corner of the map.

pgiraud:
Fixes #301

We don't have much styling options with OL3 right now.
Here's how it looks. I hope this is fine.
screenshot from 2015-12-04 17 37 17


nchauvin-brgm:
Fine to start!

PR #350, opened 2015-12-16
Revert modifications on relation between hazardcategory and admindivs

pgiraud:
This pull request reverts a commit to fix the tests broken with latest version of thinkhazard_common and pin the version to an older "working" one.


arnaud-morvan:
LGTM, I will pin thinkhazard_common version in thinkhazard_processing too.

PR #352, opened 2015-12-22
import administrative divisions

pgiraud:
With this pull request, we get a script to help importing administrative divisions using shapefiles from GAUL (fao).
The `createdb.sh` script is intented to be launched at the begining to create and populate the database.

Please review.


pgiraud:
Closing. Addressed in https://github.com/GFDRR/thinkhazard_processing/pull/32

Issue #133, opened 2015-05-21
Choose a licence


fvanderbiest:

ingenieroariel:
GPL

pgiraud:
I assume this means GPL v3. @ingenieroariel can you please confirm?

pgiraud:
@fvanderbiest Can you work on this one?

ingenieroariel:
GPL v3 is fine.

PR #383, opened 2016-01-19
GPL v3 license

fvanderbiest:
Related issue is https://github.com/GFDRR/thinkhazard/issues/133

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -1,0 +1,6 @@
+Authors ordered by first contribution.
+
+Eric Lemoine <eric.lemoine@camptocamp.com>
+François Van Der Biest <francois.vanderbiest@camptocamp.com>
+Pierre Giraud <pierre.giraud@camptocamp.com>
+Arnaud Morvan <arnaud.morvan@camptocamp.com>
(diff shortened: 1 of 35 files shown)


pgiraud:
Instead of having specifying the authors for each file, I'd use an AUTHORS.txt file.
It should be much easier to maintain.

pgiraud:
Also, the year should be 2016 not 2015.

gurneyalex:
@pgiraud you may want to use 2015-2016 if the project started last year.

Authors / Contributors can be maintained in an external file, yes. 

You also want to update setup.py with the license / copyright information. And the MANIFEST.in too (or check that the license file  and contributors files are indeed included in the tarball).

fvanderbiest:
> Also, the year should be 2016 not 2015.

The year is ```2015-``` which is supposed to mean: "from 2015 on"
Not sure it is valid though.

gurneyalex:
@fvanderbiest Does not work like that. The year is the year of publication of the version of the work. So 2016 in this case. If you make a significant change next year, you wlll be entitled to update the date to 2017. 

The year is used to find out when a version falls in the public domain according to local laws. 

As ususal::

    from std_disclaimers import ianal

pgiraud:
We can merge.

PR #386, opened 2016-01-20
Update to about.jinja2: Think Hazard! to ThinkHazard!

stufraser1:
Updated Think Hazard! to ThinkHazard!, consistent with the main branding


pgiraud:
Thanks. Merging.

PR #401, opened 2016-02-02
Remove french word "sig"

fvanderbiest:
Please review

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -80,9 +80,9 @@ Path to main data folder, example:
 
 .. code:: yaml
 
-    data_path: /var/sig
+    data_path: /tmp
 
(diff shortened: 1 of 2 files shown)


pgiraud:
Will this work on our demo server?

fvanderbiest:
Indeed, this folder has to be created by root.

fvanderbiest:
OKay, maybe it should be /tmp by default.
Everyone will expect a different location for production (eg: ```/var/www/vhosts/wb-thinkhazard/private/data``` in our case)

pgiraud:
OK for /tmp.

fvanderbiest:
Done.

PR #416, opened 2016-02-15
Return PDF as file response

pgiraud:


production.ini
```diff
@@ -28,6 +28,9 @@ node_modules = %(here)s/node_modules
 
 local_settings_path = %(here)s/local.ini
 
+# path where pdf reports are archived
+pdf_archive_path = %(here)s/.build/reports
```
fvanderbiest: Maybe we should add a note in the readme stating that this path should be customized in local.ini (or else, archived reports might disappear on `make clean`)

fvanderbiest:
LGTM

PR #424, opened 2016-02-18
Page to list hazardsets

pgiraud:
Please review.


thinkhazard/templates/hazardsets.jinja2
```diff
@@ -0,0 +1,27 @@
+<style>
```
arnaud-morvan: Could be in admin folder and inherit admin base template

arnaud-morvan:
I think you could add a pill in admin nav-bar see tamplates/admin/common.jinja2

arnaud-morvan:
Note that I have reworked nav-bar in my current work on climate changes

pgiraud:
New tab pill to be added later.

PR #425, opened 2016-02-18
Remove square brackets for associations fields

pgiraud:


pgiraud:
As discussed with @arnaud-morvan. Merging.

PR #439, opened 2016-02-24
Use jinja route_url filter everywhere. Fix #434

### LINKED ISSUE #434: make the route_url usage consistent throughout the jinja2 templates
Sometimes we use a filter, sometimes we don't.

arnaud-morvan:


pgiraud:
Please merge.

Issue #391, opened 2016-01-22
Fix map display of countries across the dateline


stufraser1:
Hi
When displaying Russian Federation in TH, the mapping shows incorrectly - with repeated parts of the globe. We believe this is due to the admin0 boundaries crossing the dateline in the east of the country.
e.g. http://wb-thinkhazard.dev.sig.cloud.camptocamp.net/beta2/wsgi/report/204/FL
Please could you have a look and fix this to display Russia properly?

Many Thanks

arnaud-morvan:
Apparently ST_Shift_Longitude (http://postgis.net/docs/manual-1.4/ST_Shift_Longitude.html) transform x=180 to x=-180 but 180 is not < 0.
This seems to be a bug in ST_Shift_Longitude.

I will propose a workaround with a translation of 180 degrees before and after applying ST_Shift_Longitude.

See this for example:

```
	SELECT
		box2d(ST_Translate(ST_Shift_Longitude(ST_Translate(geom, 180, 0)), -180, 0))
		--ST_Shift_Longitude((ST_Dump(geom)).geom) AS geom
		--ST_Translate(geom, 180, 0) AS geom
		
	FROM datamart.administrativedivision
	WHERE code = 204
```

arnaud-morvan:
Note that I did not succeed to produce a simple sql exemple that highlights the bug observed with ST_Shift_Longitude.

pgiraud:
It doesn't seem to be an issue with PostGIS and ST_Shift_Longitude.
It looks like there something wrong with the data.
I extracted the polygon for the division Cakaudrove (code 40201) and I found longitude looking like: 180.0000000000001.
This is not correctly taken into account by shift longitude.
I still don't know if this rounding issue comes from the initial shapefile, or if it happened when I used shp2pgql to import it in PostGIS.

PR #442, opened 2016-02-24
Translate before applying ST_Shift_Longitude. Relate #391

arnaud-morvan:
Workaround for ST_Shift_Longitude behavior on date line.


pgiraud:
This looks like a working workaround though I'm not happy with it. See https://github.com/GFDRR/thinkhazard/issues/391#issuecomment-188682930
Please merge.

PR #445, opened 2016-02-25
Generate PDF

pgiraud:
With this pull request, I add the support for PDF generation.
The PDF is not completely finished. Still I would like this to be merged before I continue.

Thanks for the help from @tsauerwein
https://github.com/tsauerwein/thinkhazard/commit/0a9199d55264dfddd7f6c1d7e122c09d12f493b0

Please review.


development.ini
```diff
@@ -29,6 +29,9 @@ local_settings_path = %(here)s/local.ini
 
 node_modules = %(here)s/node_modules
 
+# path where pdf reports are archived
+pdf_archive_path = %(here)s/.build/reports
```
arnaud-morvan: We could use same data_path for pdf reports and hazardset layers.
But data_path is not set by default in pyramid config for now.

thinkhazard/views/pdf.py
```diff
@@ -0,0 +1,143 @@
...
+            request.route_url(
+                'report_print',
+                divisioncode=division_code,
+                hazardtype=cat.hazardtype.mnemonic
+            )
+        )
+
+    cover_url = request.route_url('pdf_cover', divisioncode=division_code)
+
+    command = \
```
arnaud-morvan: Here he does not find wkhtmltopdf.
I remember that we had the same problem with developement.ini in processing load settings with wsgi.
Current folder is not project root. Path should be absolute ex:
```
settings = get_appsettings(os.path.join(os.path.dirname(__file__),
                                            '../../development.ini'))
```

development.ini
pgiraud: Good suggestion. Taken into account.

thinkhazard/views/pdf.py
pgiraud: Changed. Thanks.

pgiraud:
Merging now.

PR #446, opened 2016-02-25
Add basic authentification for admin ui

arnaud-morvan:
Note that with this method, authentification is needed only with WSGI, not for tests.


pgiraud:
LGTM. Thanks. Please merge.

PR #447, opened 2016-02-25
Added geonode regions of the world to the csv. This fixes #441

### LINKED ISSUE #441: Regions only include countries
With the pull request https://github.com/GFDRR/thinkhazard/pull/427, we can now link regions (as set in geonode) to GAUL administrative divisions.
In the .csv file provided by @javierdaza, only the regions that correspond to a country in GAUL are listed. We also need to link regions that are not countries to GAUL codes.

For example for Europe, we need as many lines in the CSV as countries in Europe (one for France, one for Spain, etc...)

First Header | Region id | GAUL code
------------ | -------------| --------
Europe | 5 | 85
Europe | 5 | 229
... | ... | ...

Does it make sense? Are you able to extract such information?

cc @ingenieroariel

javierdaza:
Still missing 4 regions: Channel islands, Aland islands, Saint-Barthelemey and Saint Martin.

### DIFF
--- a/data/geonode_regions_to_administrative_divisions_code.csv
+++ b/data/geonode_regions_to_administrative_divisions_code.csv
@@ -232,3 +232,238 @@
 "Yemen",251,269
 "Zambia",252,270
 "Zimbabwe",253,271
+"North America",2,46
+"West Africa",13,214
+"Pacific",256,83
+"Europe",5,104
+"Central Asia",8,250
+"Southern Africa",14,207
+"Europe",5,2647
+"Europe",5,147
+"Caribbean",255,210
+"East Africa",12,79
+"Caribbean",255,14
+"South Asia",9,231
+"Southern Africa",14,235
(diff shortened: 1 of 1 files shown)


pgiraud:
Hi. Thanks a lot for this new version.
I notice one (and only one) occurrence of "Global". Is it expected?

fvanderbiest:
Agreed with @pgiraud, there should be as many "global" lines as there are countries

ingenieroariel:
Definitely.

@javierdaza Can you check what happened and also sort the regions by name
(so that they are grouped together).

@pgiraud Should we add a fourth column with the GAUL name to help us verify?

Also "Africa", "America", "Asia" do not even apear on the list.

-a

fvanderbiest:
>  Should we add a fourth column with the GAUL name to help us verify?

Would be nice indeed.
It shouldn't break our implementation.

PR #452, opened 2016-02-26
Show level 0 & 1 in admindiv hazardsets list

pgiraud:


thinkhazard/models.py
```diff
@@ -253,6 +253,7 @@ class AdministrativeDivision(Base):
 
     leveltype = relationship(AdminLevelType)
     parent = relationship('AdministrativeDivision', uselist=False,
+                          lazy='joined', join_depth=3,
```
fvanderbiest: Are you sure 2 is not enough ?
Hopefully this will not break for administrative divisions with no parents or just a single parent.

fvanderbiest:
LGTM

PR #453, opened 2016-02-26
Redirect to technical recommendations after form submit

pgiraud:
When a user edits either a technical or climate change recommendation, the page reloads after form submit. Then the user is tempted to use the back button on his browser to get back to the list of recommendations. It may result (in Firefox for example) to showing a list which is not up-to-date, and thus confusion about whether the change has been taken into account or not.
This pull request fixes this behavior by redirecting the user directly to the list after form submit.


fvanderbiest:
OK, LGTM

PR #461, opened 2016-02-29
Geonode regions like America, Africa and Asia.

javierdaza:
Added associations with America, Africa and Asia. Also added a fourth column to help verification


pgiraud:
Thanks.

Issue #284, opened 2015-09-15
Drop-down box of placenames


stufraser1:
When entering a location name in the search box, only five options are shown; scroll through and you may never get to the one you intend to use.
E.g. If you type ‘Unit’ United States shows up, but if you only type 'Uni’ it is not on the the list, because Tunisia is returned higher in the list.  

The drop-down list would ideally show all (or at least more) locations that match the partial name that has been entered.

nchauvin-brgm:
Maybe we could increase the number of suggestions... and/or review how suggestions are ordered.

In my opinion, the first suggestions should match exactly what you are typing. For example, if you type "Uni", the suggestions should be ordered like this: 1) **Uni**ted Arab Emirates; 2) **Uni**ted Kingdom; 3) **Uni**ted States of America; 4) **Uni**ted States Virgin Islands (U.S.); 5) **Uni**ty (South Sudan). Only then, if there is no exact match anymore, suggestions should *include* the term entered, like "T**uni**sia".

Priority should also be given to full location name suggestions. I have noted an inconvenience when dealing with short location names like "Sud" in Morocco. If you type "Sud", which is a full location name, "Sud, Morocco" doesn't show up, whereas partial location names do, like "**Sud**an" or "South **Sud**an" or "Centre-**Sud**, Burkina Faso".

stufraser1:
Agreed, it makes sense that the order should be alphabetic, showing locations that begin with those letter first, then locations that contain those letters - and that it should show the admin 1/2 divisions in this logic, not after admin 0 partial matches

pgiraud:
Is it still relevant?

stufraser1:
The ordering of place names is still not sensible. Taking the 'uni' example
it still has places beginning with uni below places with uni in the middle
if the name

On Thu, 14 Jan 2016 08:48 Pierre GIRAUD <notifications@github.com> wrote:

> Closed #284 <https://github.com/GFDRR/thinkhazard/issues/284>.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/GFDRR/thinkhazard/issues/284#event-514980641>.
>
-- 
Stuart Fraser, PhD
Disaster Risk and Catastrophe Analytics Consultant
stufraser1@gmail.com
Skype: stu_fraser1
+44 (0)7455 048 044
uk.linkedin.com/in/stuartfraserdisasterrisk

pgiraud:
This looks better now for the "uni" example.
[...]

PR #469, opened 2016-03-03
Fix order of seach results on server side

arnaud-morvan:
Relate #284

### DIFF
--- a/thinkhazard/views/search.py
+++ b/thinkhazard/views/search.py
@@ -35,10 +35,10 @@ def administrativedivision(request):
     term = request.params['q']
 
     query = DBSession.query(AdministrativeDivision) \
-        .filter(AdministrativeDivision.name.ilike('%{}%'.format(term))) \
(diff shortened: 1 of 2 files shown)


thinkhazard/views/search.py
```diff
@@ -17,62 +17,31 @@
...
-             AdministrativeDivision.name.ilike(ilike))).limit(10)
-
-    data = sorted(_roundrobin(admin0s, admin1s, admin2s, limit=10),
-                  key=lambda o: o.leveltype_id)
+    term = request.params['q']
+
+    query = DBSession.query(AdministrativeDivision) \
+        .filter(AdministrativeDivision.name.ilike('%{}%'.format(term))) \
+        .order_by(
+            AdministrativeDivision.name.ilike('{}'.format(term)).desc(),
```
pgiraud: What about simply use `term` instead of `'{}'.format(term)`?

pgiraud:
Just a minor comment. Otherwise looks good.

PR #470, opened 2016-03-03
Set admindiv join depth to 2

arnaud-morvan:
According to my tests, administrative division join depth should be 2


pgiraud:
Superseded by #476.

PR #476, opened 2016-03-04
Change query to make it work with join_depth 2

pgiraud:
This PR supersedes #470.
When using a join_depth set to 2, which is better, the "administrative divisions hazardsets" in the admin UI doesn't do the job in a performant way.
With the current pull request, I fix the join_depth value, and change the query for admin page to work with the new value.
Please review.


arnaud-morvan:
I don't understand precisely, I trust you on this.

PR #479, opened 2016-03-07
Do not create apscheduler at import time

arnaud-morvan:
With scheduler created at import time, we have errors in processing scripts whereas is not used there.


pgiraud:
OK, looks good. Please merge.

PR #485, opened 2016-03-09
Don't fail with accentuated division name when getting PDF report

### LINKED ISSUE #482: PDF generation fails for divisions with accentuated characters
Test with Réunion for example.

pgiraud:
Fixes #482


arnaud-morvan:
LGTM

PR #491, opened 2016-03-11
Store processed as date

pgiraud:
With this pull request the processed column is not a boolean anymore but a date instead.
If the value is None, it means that the hazard set is not processed.


pgiraud:
Merging.

PR #501, opened 2016-03-15
Adding view for data source modal content

### LINKED ISSUE #405: Links to GeoNode in Data source
Instead of having a lightbox, let's show owner__organization with link to GeoNode detail url (distribution_url ).

pgiraud:
Fixes #405.


arnaud-morvan:
Not tested, but seems good from code view.
Note that I usually raise an HTTPNotFound when request syntax is good, but record does not exists.

PR #509, opened 2016-03-17
Use urlparse to parse sqlachemy.url in publish script. Fix #496

### LINKED ISSUE #496: Publish script doesn't work when no port is set
In the url to database, port may not be set. Default port is used.
example: `postgresql://www-data:www-data@localhost/thinkhazard`
The publish fails in this case.

arnaud-morvan:


pgiraud:
Excellent. Please merge.

PR #510, opened 2016-03-17
Use config.include route_prefix parameter

arnaud-morvan:


thinkhazard/__init__.py
```diff
@@ -73,7 +73,7 @@ def main(global_config, **settings):
         config.add_route('admin_hazardsets', '/hazardsets')
         config.add_route('admin_hazardset', '/hazardset/{hazardset}')
 
-        add_public_routes(config, prefix='/preview')
+        config.include(add_public_routes, route_prefix='/users')
```
pgiraud: "users" ?

arnaud-morvan:
Good catch, note that this has a side effect : preview index ends with a /

pgiraud:
> Good catch, note that this has a side effect : preview index ends with a /

This can be managed by apache configuration if required.

pgiraud:
LGTM. Please merge.

PR #514, opened 2016-03-17
Updated to add a new FAQ

stufraser1:
New FAQ directs users to a sheet that shows the comparison of hazard level given in ThinkHazard! and the comparison with expert expectations.
It also corrects the direction for users to give feedback via the feedback button, rather than the FAQ page


pgiraud:
Will deploy this tomorrow.

ingenieroariel:
:+1:  Go. Stu. Go.

PR #519, opened 2016-03-18
Fix add_job mock. Fix #508

### LINKED ISSUE #508: Tests never stop
Even when tests are OK, we must stop process manually.

arnaud-morvan:


pgiraud:
Please merge.

PR #531, opened 2016-03-24
Default db name is thinkhazard_processing

pgiraud:
[WIP] the documentation procedure needs more work. It should be easier for new users to understand what happens and how to get an application working.
For example, the first populated database is the admin/processing one. The url is "/admin". Then when publish the public interface is available. Not before.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -27,8 +27,8 @@ Create a Python virtual environment and install the project into it::
 
 Create a database::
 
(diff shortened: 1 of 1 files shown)


pgiraud:
Also it's not obvious that css needs to be build beforehand.

pgiraud:
Already fixed.

PR #532, opened 2016-03-24
apache conf enhancements

pgiraud:
The main idea is to get the application running at the domain root. No extra "main/wsgi".
Please review.


arnaud-morvan:
I note that we could have url overlapping between main and other instance. Apache behavior in this case should be tested.
Looks good otherwise.

pgiraud:
My tests on the dev server show that there's no overlap. Other instances still continue to work nicely.

PR #538, opened 2016-03-30
Store PDFs in YYYY/MM subdirectories

### LINKED ISSUE #534: store pdf reports in subdirectories
For example we could use `YYYY/MM/tartanpion.pdf`.

pgiraud:
Prevents filesystem overload

Fixes #534.

@arnaud-morvan Can you please review?


arnaud-morvan:
LGTM

PR #557, opened 2016-03-31
Restart apache after publication

arnaud-morvan:
We need to clear wsgi variables after publication.


pgiraud:

PR #563, opened 2016-04-01
Add support for alembic migration scripts

pgiraud:
Please review.


arnaud-morvan:
It could be possible to upgrade the admin and public database at same time.

alembic/env.py
```diff
@@ -0,0 +1,128 @@
...
+def run_migrations_online():
+    """Run migrations in 'online' mode.
+
+    In this scenario we need to create an Engine
+    and associate a connection with the context.
+
+    """
+    # use the 2-phase protocol to make sure that the migration is applied
+    # correctly to all databases or none.
+    use_two_phase = context.get_x_argument(
```
pgiraud: This is interpreted as a String (False) unfortunately.

pgiraud:
Thanks for your help @tsauerwein.
Merging.

PR #564, opened 2016-04-01
Use app init time as last_modified header

arnaud-morvan:


pgiraud:
Can you please rebase?

arnaud-morvan:
Done

pgiraud:

pgiraud:
This was actually a bad idea since we can't ensure that the init time is shared between all the wsgi processes. This results in resources being randomly considered as modified or not depending on which processed is used.
Reverting.

Issue #556, opened 2016-03-31
Make sure an error on processing a dataset doesn't stop processing for other datasets


pgiraud:

pgiraud:
It's probably due to MemoryError which is probably due to raster matrices not matching.
So we should first try to detect that matrices don't match and raise error in this case.

PR #567, opened 2016-04-01
Verify that rasters properties match during complete.

arnaud-morvan:
Relate #556

I've add a commit here to ensure that polygon and raster intersects. This should fix the memory error.


pgiraud:

Issue #547, opened 2016-03-31
Create release and put a link on the about page


pgiraud:

pgiraud:
Fixed via #579.

PR #579, opened 2016-04-05
Show application version in about page

pgiraud:
Works with "git describe" and should show tags when annotated tags are
available.

### DIFF
--- a/development.ini
+++ b/development.ini
@@ -30,6 +30,7 @@ jinja2.filters =
     route_url = pyramid_jinja2.filters:route_url_filter
     static_url = pyramid_jinja2.filters:static_url_filter
     markdown = thinkhazard.filters:markdown_filter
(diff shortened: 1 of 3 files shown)


pgiraud:
Fixes #547

arnaud-morvan:
I don't really understand why you make a filter for this. This could be executed in about view.

pgiraud:
> I don't really understand why you make a filter for this. This could be executed in about view.

Because I don't know how I can do the following in a jinja2 template:
```python
import subprocess
label = subprocess.check_output(['git', 'describe', '--always'])
```

arnaud-morvan:
I was thinking to call git describe in view and return the label to the template.

pgiraud:
It's not only one view.

arnaud-morvan:
Thanks, it looks good to me now.

PR #581, opened 2016-04-07
alembic fix

pgiraud:
With this pull request I address the following issue.

When instances are not specific, integrator usually don't set the database url in `local.ini` since the default values are already correct and don't need to be overwritten.
In this case, the `alembic.ini` was taken into account. This raised two problems:
 - race condition of config reading made so the admin database was taken into account for both admin and public applications,
 - the urls in `alembic.ini` were not correctly set.


pgiraud:
Merging.

PR #583, opened 2016-04-07
Get git version (new strategy)

pgiraud:
Instead of executing "git describe" each time the page is displayed (which doesn't work correctly in apache mod_wsgi), we put the revision number in an ".ini" and load it via the settings.


pgiraud:
Merging.

PR #585, opened 2016-04-07
Replace completed by complete in harvesting. Fix #553

### LINKED ISSUE #553: Hazard set update date doesn't seem to be updated adequately
For example, it looks like when layers from a dataset are updated, the complete is still true and updated date on the hazard set is not modified.

arnaud-morvan:


pgiraud:

PR #589, opened 2016-04-07
Archive database backups.

### LINKED ISSUE #530: Archive database
Each time the data is published, the public database should be archived.
We need to define how the archive should be done.
Creating a dump each time the data is published doesn't seem to good solution since 95% of the data is similar from one version to an other. Instead, we should investigate the usage of `barman`.

arnaud-morvan:
Fix #530

Note that developpers should have to set local ``backup_path`` setting in ``adminapp``.

### DIFF
--- a/thinkhazard/scripts/publish.py
+++ b/thinkhazard/scripts/publish.py
@@ -61,10 +61,15 @@ def main(argv=sys.argv):
     options = parse_vars(argv[2:])
     setup_logging(config_uri)
 
+    admin_database = database_name(config_uri, 'admin', options=options)
+    public_database = database_name(config_uri, 'public', options=options)
+
+    print 'Lock public application in maintenance mode'
     with open(lock_file, 'w') as f:
         f.write('This file sets the public application in maintenance mode.')
 
     # Create new publication in admin database
+    print 'Log event to publication table in', admin_database
     settings = get_appsettings(config_uri, name='admin', options=options)
     load_local_settings(settings, 'admin')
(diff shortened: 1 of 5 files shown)


arnaud-morvan:
@pgiraud : Have you a precise idea about path ?
Exemple backup path : /tmp/thinkhazard.2016-04-07T15:26:32.824867.backup
This does not take care about instanceid.
Maybe we could add a setting to disable this outside of production server ?

pgiraud:
Directory to put the archives in is `/srv/archives`. We could add a `backup` subfolder not to mix things with generated reports.
The path should be configurable (as you already did).

I would say that if the variable is not set, no backup should be done. Let's talk about the variable to documentation and remove it (or comment it) from dev/prod.ini files.

thinkhazard/scripts/publish.py
```diff
@@ -75,6 +76,17 @@ def main(argv=sys.argv):
     admin_database = database_name(config_uri, 'admin', options=options)
     public_database = database_name(config_uri, 'public', options=options)
 
+    backup_path = os.path.join(
+        settings['backup_path'],
+        'thinkhazard.{}.backup'.format(datetime.utcnow().isoformat()))
+    print backup_path
```
pgiraud: Do be removed.

arnaud-morvan:
I've added some outputs, and lightly refactored .ini files for lisibility (easy compare).

We need a path for backup, because we need to do create the file. A nice to have could be a setting to magically remove the backup in development mode.

pgiraud:
LGTM. Merging.

PR #591, opened 2016-04-08
Make sure outputs are removed when complete is false

### LINKED ISSUE #590: Outputs should be deleted when hazard set is marked as not processed
If a hazard set was "processed" and set to "not processed", the corresponding outputs should be removed in order not to be taken into account later.

pgiraud:
Fixes #590.

### DIFF
--- a/thinkhazard/processing/completing.py
+++ b/thinkhazard/processing/completing.py
@@ -27,6 +27,7 @@ from ..models import (
     DBSession,
     HazardSet,
     Layer,
+    Output,
     )
 
 from ..processing import BaseProcessor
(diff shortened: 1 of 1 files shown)


arnaud-morvan:
I note that this was already done here : 
https://github.com/GFDRR/thinkhazard/blob/pgiraud_clean_output/thinkhazard/processing/processing.py#L99:L103

Here, I think we should only put processed to False because of this:
https://github.com/GFDRR/thinkhazard/blob/pgiraud_clean_output/thinkhazard/processing/harvesting.py#L482
(in case the quality change, we want to complete but not process)

pgiraud:
There's actually an other case. When for example the processing parameters change and the hazard set becomes incomplete.

PR #592, opened 2016-04-08
Add detailed text for technical recommendation

### LINKED ISSUE #546: Detailed recommendation
Additional detailed text should be available for users for each technical recommendation.
A "more information" link should be added in the UI. Clicking on the link should open a modal window.

pgiraud:
Fixes #546


arnaud-morvan:
I note that it should be good to have a title in the dialog, for exemple "Recommendation details" for visual apperance.
I see a makdown filter. I think this should be documented at some time.

alembic/versions/f58bd58cd6d4_.py
```diff
@@ -0,0 +1,29 @@
+"""empty message
```
arnaud-morvan: This file should have an explicit name

pgiraud:
For the visual appearance, I think that the place holder is sufficient.

PR #595, opened 2016-04-08
Changing drought to water scarcity

### LINKED ISSUE #545: Change "drought" to "water scarcity"
Only change the label in the interface.

pgiraud:
Fixes #545.


arnaud-morvan:
Tested and works well for me.
Will conflict with #596 on alembic down_revision.

PR #607, opened 2016-04-09
Make sure HazardSet was already created

pgiraud:


thinkhazard/models.py
```diff
@@ -477,7 +478,7 @@ def layer_by_level(self, level):
 
 @listens_for(HazardSet.processed, 'set')
 def on_hazardset_processed_set(target, value, oldvalue, initiator):
-    if value is None:
+    if value is None and value != oldvalue and oldvalue is not NO_VALUE:
```
arnaud-morvan: Is it not possible to use ``target.id is not None`` here ?

pgiraud:
Fixes in #604.

PR #611, opened 2016-04-11
No autoflush in delete outputs

arnaud-morvan:


pgiraud:
This pull request fixes issues encountered when the layers are created for the first and thus the corresponding hazard set is created as well.

pgiraud:

PR #614, opened 2016-04-11
Don't show mask return period in hazardset data source info

### LINKED ISSUE #613: Mask return period should not be displayed source modal window
screenshot from 2016-04-11 13 36 51

pgiraud:
Fixes #613.


pgiraud:
Trivial. Merging.

PR #624, opened 2016-04-12
Do not use mask layer for hazardset quality calculation

### LINKED ISSUE #618: Data Quality should not be read from mask layers
Flood mask layer have data quality currently set to 0,0.
In the admin hazardsets page, data quality rating of layers with a mask show as 0,0, but in geonode the values are 5 (calculation method) and 2 (scientific).

It seems the admin page is incorrectly showing the data quality from mask layer, not RP layer.

It seems the RP layer data quality values are used in the decision tree, as Alfieri is used correctly for France (If values 0,0 were being used, data source would show as GAR15)

arnaud-morvan:
Fix #618.

To update an existing database, execute : 
```
UPDATE processing.hazardset SET complete = FALSE
```
and ``make complete``.

This will update quality fields in hazardset table.

### DIFF
--- a/thinkhazard/processing/completing.py
+++ b/thinkhazard/processing/completing.py
@@ -131,6 +131,7 @@ class Completer(BaseProcessor):
             func.min(Layer.calculation_method_quality),
             func.min(Layer.scientific_quality)) \
             .filter(Layer.hazardset_id == hazardset.id) \
+            .filter(Layer.mask.isnot(True)) \
             .group_by(Layer.local)
 
         if stats.count() > 1:


thinkhazard/processing/completing.py
```diff
@@ -131,6 +131,7 @@ def complete_hazardset(self, hazardset_id, dry_run=False):
             func.min(Layer.calculation_method_quality),
             func.min(Layer.scientific_quality)) \
             .filter(Layer.hazardset_id == hazardset.id) \
+            .filter(Layer.mask == false()) \
```
pgiraud: Why not simply using `Layer.mask.is_(False)` or `Layer.mask.isnot(True)`?

arnaud-morvan: I don't mind, no specific reason, you say it's simpler, it seems equal for me.

pgiraud: No need to import anything.

pgiraud:

PR #635, opened 2016-04-25
Fix errors when restarting postgreSQL

pgiraud:
Happens on recent Debian boxes:

```
sudo -u postgres /etc/init.d/postgresql restart
Restarting postgresql (via systemctl): postgresql.serviceFailed to get D-Bus connection: No such file or directory
 failed!
```


arnaud-morvan:
LGTM, all servers concerned by this project will support it.

PR #638, opened 2016-04-27
Use return period from 2 to 5 for river flood mask layer

arnaud-morvan:
And use threshold from mask unit for mask layer


pgiraud:
LGTM. Merging.

PR #716, opened 2017-02-02
Update divisions

pgiraud:
This pull request's goal is to update the administrative divisions:
 - some divisions names have been updated,
 - names in French and Spanish have been added for level 0 when available,
 - `import_divs` script has been updated to import directly from shapefile,
 - some divisions have change (Kosovo for example).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -81,15 +81,19 @@ initdb_force:
 
 .PHONY: import_admindivs
 import_admindivs: .build/requirements.timestamp \
-		/tmp/thinkhazard/admindiv/$(DATA)/g2015_2014_0.sql \
-		/tmp/thinkhazard/admindiv/$(DATA)/g2015_2014_1.sql \
-		/tmp/thinkhazard/admindiv/$(DATA)/g2015_2014_2.sql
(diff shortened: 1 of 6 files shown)


pgiraud:
@stufraser1 
I finally managed to get something working to update the administrative divisions. It took more work than expected. The good thing is that it was worth since it was required for the names translations.

I invite you to go the integration instance and check if everything works as expected.
http://int.thinkhazard.org/

This instance runs a copy of the production database with the update.
I manually change (via the admin interface) the climate change recommendation for Landslide so that it applies for Kosovo as well.

Once you think this works as expected, I'll deploy the code and database on the production server.

stufraser1:
I'm currently unable to view http://int.thinkhazard.org/ due to 'internal server error'

pgiraud:
Can you please try again?

stufraser1:
Kosovo works well. Please deploy as we need to get that live ASAP.

pgiraud:
Production is up-to-date!

stufraser1:
Thank you!

PR #720, opened 2017-02-10
include licence statement in v1 about page

stufraser1:
@pgiraud 
Please include this licence statement in v1 about page. I edited the about_body.jinja file (my fork)
ThinkHazard! is available under the GNU General Public Licence, Version 3, 29 June 2007. Text content is licenced under CC-BY-SA. Classified hazard levels are licenced under CC-BY. Original hazard data are licenced under their original terms, which are contained in the associated layer metadata.


pgiraud:
If it needs to be done for v1, then it should be done against 1.5 branch (last version of v1).
I created #721 to address this.

pgiraud:
@stufraser1 This has been fixed and deployed.

PR #727, opened 2017-02-16
Put language in URLs

pgiraud:


pgiraud:
I added a test to check that language is correctly set in the page (HTML `lang` attribute).
I also added some code to automatically redirect to default language if language set in URL is not available.
Merging.

PR #731, opened 2017-02-22
Make the polygons bigger if they are really small

### LINKED ISSUE #655: Need to refine mapping of small islands
Improvements in mapping small islands required in next version; see for example
http://thinkhazard.org/report/2278-palau-administrative-unit-not-available
which is too difficult to read the mapped levels

This effect is due to the large distances between admin units.

Known locations affected: 
Palau
Maldives (no islands visible at all)

Possible solution:
- Add more flexible zoom function to highlight a square zoom area? All admin units that intersect zoom area, to be shown with classified hazard level?

pgiraud:
Fixes #655.
Interactivity enhanced by making the really small and scattered islands a bit bigger than they actually are.

Demo
http://wb-thinkhazard.dev.sig.cloud.camptocamp.net/small_islands/en/report/154-maldives/CY
http://wb-thinkhazard.dev.sig.cloud.camptocamp.net/small_islands/en/report/189-palau/CY

@stufraser1 Can you please validate this?

### DIFF
--- a/thinkhazard/static/js/report.js
+++ b/thinkhazard/static/js/report.js
@@ -37,9 +37,13 @@
 
   var tooltipEl = $('#map .map-tooltip');
 
+  var hitTolerance = 3;
+
   // change mouse cursor when over division
   map.on('pointermove', function(e) {
-    var feature = map.forEachFeatureAtPixel(e.pixel, filterFn);
+    var feature = map.forEachFeatureAtPixel(e.pixel, filterFn, {
+      hitTolerance: hitTolerance
+    });
     var cursor = '';
 
     tooltipEl.empty().removeClass('neighbour').hide();
@@ -76,7 +80,9 @@
 
   // drill down
   map.on('click', function(e) {
-    var feature = map.forEachFeatureAtPixel(e.pixel, filterFn);
+    var feature = map.forEachFeatureAtPixel(e.pixel, filterFn, {
+      hitTolerance: hitTolerance
+    });
     if (feature) {
       var url = feature.get('url');
       if (app.hazardType) {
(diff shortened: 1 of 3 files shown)


stufraser1:
I think this works well. what is the mechanism you've used? I see a buffer has been applied to maldives, but the extent for Palau has changed in addition to the buffer, so how is this done and can it be replicated for other islands? We will need to include the explanation in an FAQ / method document to clarify any queries on why islands look bigger than they are. Good solution though

pgiraud:
The method is:
 - determine extent of the division,
 - zoom the map to the given extent and compute the resolution of the map for this zoom level,
 - compute the area of the division (sum of areas of the islands),
 - if the area is lower than a empiric threshold (based on resolution) then add a buffer (based on resolution) around the geometry,
 - otherwise don't touch the geometry

For Palau, in the current production version, small islands (in the South-West corner) are taken into account to display the map but are not shown because simplification remove them. With the "buffer" fix, tiny islands show up again. Extent doesn't look the same but actually is.

What other islands are you talking about?

stufraser1:
OK, thank you. Could also apply to e.g,. American Samoa, Marshall Islands. Essentially I most want to know that the process is automated. It seems so, including in the selection of the territory to apply this method on.

thinkhazard/static/js/report.js
```diff
@@ -39,7 +39,9 @@
 
   // change mouse cursor when over division
   map.on('pointermove', function(e) {
-    var feature = map.forEachFeatureAtPixel(e.pixel, filterFn);
+    var feature = map.forEachFeatureAtPixel(e.pixel, filterFn, {
+      hitTolerance: 3
```
tonio: Would be better to set this value in a single place, rather than repeating 3 times.

PR #732, opened 2017-02-22
Manage which hazard types should be shown

pgiraud:
Remove Air Pollution from UI

### DIFF
--- a/thinkhazard/static/less/report.less
+++ b/thinkhazard/static/less/report.less
@@ -122,7 +122,7 @@ body {
   transform-origin: right top;
   -webkit-transform-origin: right top;
   -ms-transform-origin: right top;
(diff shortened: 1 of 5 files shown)


thinkhazard/static/less/report.less
```diff
@@ -122,7 +122,7 @@ body {
   transform-origin: right top;
   -webkit-transform-origin: right top;
   -ms-transform-origin: right top;
-  width: ~"calc(37.7em + 7px)";
+  width: ~"calc(34em + 7px)";
```
tonio: Those magic numbers should be documented.

pgiraud:
@stufraser1 integration instance updated.
http://int.thinkhazard.org/en/ (Pay attention to possible cache issues)

Issue #652, opened 2016-05-13
Repeated data sources in UI


stufraser1:
Beneath the map, for Kenya and Uganda EQ, data source 'Impact Forecasting' is stated more than once, though only one data source is used

e.g. http://thinkhazard.org/admin/preview/report/133-kenya/EQ

pgiraud:
The data sources that are displayed at a country level may be the concatenation of several sources in case the sources for lower levels are different from one division to an other.
In the case of Kenya, there's at least one subdivision for which the data set used is the one for Uganda.
For example, the EQ  level for Turkana division (Kenya's subdivision) is determined using the EQ-UG-IF dataset.
http://thinkhazard.org/admin/preview/report/51392-kenya-rift-valley-turkana/EQ

This most probably happens because there's no direct link between a dataset and the division it's supposed to take into account.

pgiraud:
Did my comment from May last year make sense?
For the Kenya we have two different datasets (EQ-UG-IF and EQ-KE-IF) that are taken into account.
Do you have any opinion on what we should do with such a use case?

stufraser1:
In this case, the Data Quality ratings are the same, and they are both local. The preference would be to use Kenya only in this case. 
I understand why this issue occurs: there is at least one ADM2 unit that is covered by both UGA and KEN datasets. KEN should be used, but for some reason (due to the order in which datasets are searched while processing?) UGA is already assigned and used before KEN can be considered. As the DQ are the same KEN, is not considered to override UGA. Is there any mechanism that can be built in to solve this?
In cases where two datasets cover the same unit and differ in DQ, the higher should be used. I understand that this does happen correctly.

stufraser1:
We identify the region / country that the data applies to in the metadata under 'regions'. Can we use that field to restrict the data that is used. E.g., for Kenya, only data that has regions defined as 'Kenya' or its parent regions ('East Africa', 'Africa' or 'global') could be considered. This might require a lookup between ADM names used in Geonode and GAUL, for a few countries.

pgiraud:
Good point! Thanks @stufraser1 
It looks like we already have the lookup table handy. This was already provided by Javier in order to link documents with administrative divisions. We already have information on how to admin divs relate to regions in the DB.

Note for Camptocamp: what we need to do is to fetch "/api/layers/{id}" and extract the regions from there. Create a new table to link layers to regions in the DB. Make it so the processing on applies to the divisions corresponding to the associated regions.

[...]

PR #739, opened 2017-03-02
Take hazardset regions into account when processing

pgiraud:


pgiraud:
This should fix #652.
Also this should improve the processing performances by reducing the amount of administrative divisions taken into account for each hazard set.
This assumes that the regions for the hazard sets are defined in Geonode. If no region is set, then the hazard set will not be taken into account for a single division.

@stufraser1 I'll set a demo up today so that you can confirm that it works correctly.

stufraser1:
Can this error be recorded in the administration UI if it occurs, preventing upload to TH? As we do for missing layers

pgiraud:
Demo:
http://wb-thinkhazard.dev.sig.cloud.camptocamp.net/regions/admin/preview/en/report/133-kenya/EQ

On this demo only the EQ-UG-IF and EQ-KE-IF hazard sets are processed.

@stufraser1 can you please validate?

PR #758, opened 2017-05-26
Update thinkhazard_processing.yaml

stufraser1:
Added new WF (wildfire) RP and intensity thresholds for processing set of intensity rasters.
Added new UF (urban_flood) values (preprocessed layer, values 1-4)
Added new RF (river_flood) values (preprocessed layer, values 1-4), changed existing RF codes to be void (but would like to retain these for possible future use). Please check whether this changing of the code invalidates it sufficiently, or whether it needs to be removed.

### DIFF
--- a/development.ini
+++ b/development.ini
@@ -48,6 +48,7 @@ jinja2.newstyle = true
 node_modules = %(here)s/node_modules
 
 local_settings_path = %(here)s/local.ini
+processing_settings_path = %(here)s/thinkhazard_processing.yaml
 
 # By default, the toolbar only appears for clients from IP addresses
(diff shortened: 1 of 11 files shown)


arnaud-morvan:
I've rebased on top of master to get travis happy. Concerning the invalidation of hazardset, note that we will completely clear the cache to get all data from new server and completely re-process everything.

arnaud-morvan:
I've fixed tests by pinning old settings for tests, because changing some values in processing settings should not break unit tests.

@fvanderbiest : can you take a look ?

fvanderbiest:
LGTM

PR #767, opened 2017-06-15
Change style for link on map

tonio:
screenshot from 2017-06-15 10-39-25

Thin black line could be confusing, what about a dashed one?

### DIFF
--- a/thinkhazard/static/js/report.js
+++ b/thinkhazard/static/js/report.js
@@ -276,8 +276,10 @@
       })
     ];
     var styleFn = function(feature) {
-      var strokeColor = feature.get('neighbour') ? '#337ab7' : '#000000';
+      var strokeColor = feature.get('neighbour') ? '#333' : '#000000';
       strokeStyle.setColor(strokeColor);
+      var strokeWidth = feature.get('neighbour') ? 1 : 2;
+      strokeStyle.setWidth(strokeWidth);
       return styles;
     };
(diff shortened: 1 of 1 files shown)


stufraser1:
Dashed lines should be reserved for disputed borders. Didn't we have disputed borders represented as dashed lines in a previous dev version? If so this border style should be reinstated. Example of disputed border is jammu and kashmir. Gaul files have a field stating whether an adm unit is disputed or not. Please confirm whether you have a previous dev version / basemap in v2 dev that included such borders as dashes.

tonio:
That was previous style for disputed borders:
screenshot from 2017-06-19 15-07-11

It seems to have disappear in the new one:
screenshot from 2017-06-19 15-07-41

I can change «Go to» links to what was requested first, will look like: 
screenshot from 2017-06-19 15-04-36

fjacon:
@stufraser1 :

- is Antoine's proposal OK ? (3rd picture)
- For disputed boundaries, the mapbox style has to be modified.

stufraser1:
Antoines proposal looks ok. I still have some concern that the non select Ted boundaries look like rivers, but maybe this will appear ok when I add in boundaries to the basemap. I'll have to supply a new basemap later this week

fjacon:
@tonio so you can go ahead with merging this PR.

tonio:
Deployed on integration

PR #769, opened 2017-06-15
Add api key to access geonode api

Vampouille:


Vampouille:
We need to fix unit tests as it required `username` and `api_key`

PR #819, opened 2017-10-24
Update geonode_regions_to_administrative_divisions_code.csv

werpish2:
Changed Czech republic to Czechia in all instances.


stufraser1:
PR will not be merged - we should retain official country naming of the World Bank for this product. This PR should not be merged.

PR #844, opened 2018-04-10
Update report.jinja2

stufraser1:
Updated hazard method box text, for peer review in int.


tonio:
I pushed those commits into PR #843

PR #871, opened 2020-02-04
Python3 migration

tonio:
I created the [before_py3_docker](https://github.com/GFDRR/thinkhazard/releases/tag/before_py3_docker) tag to keep reference to repo state before this work.

But I think it will be easier to work with master now. Py3 migration is completed, tests are broken until docker migration is completed.


dev-requirements.txt
```diff
@@ -1,8 +1,8 @@
 #
 # This file defines the requirements.txt for development.
 #
-flake8==2.2.2
-pep8-naming==0.2.2
-nosier==1.1
+flake8
+pep8-naming
+nosier
```
tkohr: Doesn't it risk to break the app to remove pinned versions?

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:3.8
+FROM python:3.8-slim
```
tkohr: The image does not include curl/wget which we might need to install node, postgres-client later. Not sure if it's a better practice to take a larger image or install more stuff by hand. Might be adapted again in multi stage build anyways.

dev-requirements.txt
tonio: That was needed for the upgrade, we have to keep in mind on pinning it again at the end of developments.

Dockerfile
tonio: Front build will be done in another image, let's see later for the postgres-client.

PR #872, opened 2020-02-04
Dockerize print using pyppeteer

tonio:
~~Based on https://github.com/GFDRR/thinkhazard/pull/871~~

Synchronous request for now, may be enough with the performance gain of using pyppeteer. We'll see later if asynchronous generation needs to be restored.

### DIFF
--- a/thinkhazard/static/js/report.js
+++ b/thinkhazard/static/js/report.js
@@ -509,9 +509,10 @@
     }
     map.updateSize();
   };
-  mediaCheck();
-
-  $(window).on('resize', mediaCheck);
+  if (!app.pdf) {
+    mediaCheck();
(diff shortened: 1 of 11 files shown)


thinkhazard/static/js/report.js
```diff
@@ -317,39 +317,27 @@
...
-      .done(function(data) {
-        btnStatus(true);
-        checkPdfStatus(data.report_id);
-      })
-      .fail(function() {
-        alert("Something went wrong");
-        btnStatus(false);
-      });
+    btnStatus(true);
+    fetch(app.createPdfReportUrl, { method: 'POST'})
```
arnaud-morvan: we could use GET request here

PR #877, opened 2020-02-20
[GSWBTH-26] Makes use of S3 in publish script and PDF printing

tkohr:
PR is ready for review.

TODOS to be done in separate tasks (depend on openshift deployment):
* set maintenance mode (publish.py)
* clear cached data (publish.py)


thinkhazard/views/pdf.py
```diff
@@ -145,11 +150,11 @@ def create_pdf_report(request):
     division_code = request.matchdict.get("divisioncode")
 
     base_path = request.registry.settings.get("pdf_archive_path")
-    report_id = _get_report_id(division_code, base_path)
-
-    file_name = _get_report_filename(base_path, division_code, report_id)
+    object_name = "{:s}.pdf".format(division_code)
```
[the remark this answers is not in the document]
tkohr: That's right, I somehow thought we would clear all reports on s3 when data is updated in thinkhazard. Sorry, but I have a little doubt about the logic then. Don't I have to change the current logic
- try to download report from s3 to serve to user
- if report does not exist on s3, create and upload it to s3
to
- try to download report from s3 to serve to user
- in any case create and upload report to s3 (and overwrite existing, which would also need the same name without publication date in fact, wouldn't it?)

docker-compose.yaml
```diff
@@ -19,3 +19,26 @@ services:
...
     ports:
       - 8080:6543
+
+  minio:
+    image: minio/minio
+    ports:
+      - "9000:9000"
+    command: 'server start'
+
+  # minio-client:
```
tkohr: @arnaud-morvan @tonio  I don't get minio-client to run after minio (server) even though using 'depends_on'. Is it because entrypoint does not wait for the container to be ready? Any idea how to handle this?

It means currently thinkhazard bucket is not created automatically for dev, but must be created via minio webinterface.

docker-compose.yaml
```diff
@@ -19,3 +19,26 @@ services:
     command: 'pserve c2c://${INI_FILE}'
     ports:
       - 8080:6543
+
+  minio:
```
tkohr: Should I declare a volume for 'minio' or isn't it necessary since we don't need a mapping to the host?

thinkhazard/views/pdf.py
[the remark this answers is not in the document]
tkohr: Ok got it pub date and locale are added to s3 object

Makefile
```diff
@@ -3,6 +3,7 @@ AUTHUSERFILE ?= /var/www/vhosts/wb-thinkhazard/conf/.htpasswd
 DATA ?= world
 
 -include local.mk
+-include .env
```
tkohr: @arnaud-morvan I guess this is not the right approach. How can I pass var envs to publish target without including .env here?

arnaud-morvan: As `.env` is generated through `Makefile`, if you use it in `Makefile` it is a dependency loop.
I do not know why you want to do this, but for sure it is not the right way.

Variables in .env are passed to the containers so they are accessible :
- in publish.py or any python script with `os.environ['MYVAR']`
- in bash with `$MYVAR`.

tkohr: @arnaud-morvan Yes that makes sense. Maybe the problem lies in my target (https://github.com/GFDRR/thinkhazard/blob/8a2b3d29ceff5b1ce70170e7ec59713da213715c/Makefile#L161), but I always get the values that are exported in the Makefile as env vars in the container and not the ones in declared in .env

Makefile
```diff
@@ -9,6 +9,7 @@ else
 endif
 AUTHUSERFILE ?= /var/www/vhosts/wb-thinkhazard/conf/.htpasswd
 DATA ?= world
+INI_FILE ?= development.ini
```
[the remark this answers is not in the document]
tkohr: ok

Makefile
tkohr: Ok we just clarified that the behavior is as expected, thanks! Publish for other values than default values needs to be called with var env file currently (which will change in openshift deployment).

PR #881, opened 2020-03-05
[GSWBTH-10] Tracks API usage statistics

tkohr:
PR sends hits to GA to track API usage (when production.ini is used).

A custom report is set up in GA interface to display this data.

### DIFF
--- a/thinkhazard/views/api.py
+++ b/thinkhazard/views/api.py
@@ -23,11 +23,14 @@ from pyramid.httpexceptions import HTTPNotFound
 from thinkhazard.models import HazardCategory, HazardType, HazardLevel
 from thinkhazard.views.admin import admindiv_hazardsets_hazardtype
 
+from ..analytics import GoogleAnalytics
+
 
 @view_config(route_name="api_hazardcategory", renderer="json")
 def api_hazardcategory(request):
     hazard_type = request.matchdict["hazard_type"]
     hazard_level = request.matchdict["hazard_level"]
+    GoogleAnalytics().hit(request.path)
 
     try:
         hazard_category = (
@@ -47,4 +50,5 @@ def api_hazardcategory(request):
 @view_config(route_name="api_admindiv_hazardsets_hazardtype", renderer="json")
 def api_admindiv_hazardsets_hazardtype(request):
     data = admindiv_hazardsets_hazardtype(request)
+    GoogleAnalytics().hit(request.path)
     return data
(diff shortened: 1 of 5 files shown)


thinkhazard/views/api.py
```diff
@@ -24,11 +24,14 @@
...
 from ..models import DBSession, HazardCategory, HazardType, HazardLevel
 
+from ..analytics import GoogleAnalytics
+
 
 @view_config(route_name="api_hazardcategory", renderer="json")
 def api_hazardcategory(request):
     hazard_type = request.matchdict["hazard_type"]
     hazard_level = request.matchdict["hazard_level"]
+    GoogleAnalytics().hit(request.path)
```
tonio: I think GA request will be blocking that way.
You may take a look at https://github.com/ross/requests-futures to have non-blocking requests.
(or maybe @arnaud-morvan will have other options)

tkohr: Ok thanks. Do you mean blocking in the way that the code is waiting for the response or that GA does not take the request into account? I still have the impression that the code is executed, as the API continues to serve the response when the hit is sent to GA.

tonio: Yes, I think it will wait for GA answer before returning API response. So if GA is down API too (unlikely). But should slow down API response a bit.

tkohr: Thanks I used requests-futures to make request async after trying with asyncio

tkohr:
> Reading `requests-futures` doc, if you call `.result()` here you're making it synchronous again.

Ah really !? I didn't get that on the first read. I will get rid of it as google will rarely not respond with a status 200 and the status is not even a confirmation if the hit is valid / counted.

tonio:
```
  File "/app/thinkhazard/views/report.py", line 264, in report_json
    GoogleAnalytics().hit(request.path, "report-division_code-hazard_type")
  File "/app/thinkhazard/analytics.py", line 29, in __init__
    self.debug = "" if os.environ['INI_FILE'] == "production.ini" else "/debug"
  File "/usr/local/lib/python3.8/os.py", line 675, in __getitem__
    raise KeyError(key) from None
KeyError: 'INI_FILE'
```
Travis fail.

tonio:
Looks like travis doesn't need to be updated if #882 is merged. Maybe @arnaud-morvan can merge it if he's happy with, and you can rebase this one.

PR #888, opened 2020-03-26
Push to docker hub and transifex

arnaud-morvan:


arnaud-morvan:
Note that I run the tests and checks between build and push.
And I like the idea that we can push from workstation with the same script.

PR #891, opened 2020-03-30
Delete layer from cache if data_lastupdated_date has changed

arnaud-morvan:
Because new file may have the same name as old one.


arnaud-morvan:
@tkohr : yes for sure this should be correctly handled on master, but I do not know precisely how we will handle cache on master / open Shift environment.
We will take care of this when merging 2.0 in master.

PR #892, opened 2020-04-01
Temporary overrides

arnaud-morvan:
Temporary overrides to publish landslide as soon as possible.


tonio:
Tests are failing due to crs check.

arnaud-morvan:
Yes, I made a little test data refactorization and added a test for bad crs.

PR #893, opened 2020-04-01
Push to docker hub and transifex (using docker action)

arnaud-morvan:
Same as #888 but using docker/build-push-action@v1


arnaud-morvan:
#888 has win

PR #894, opened 2020-04-02
Deletes non-harvested layers and hazardsets

tkohr:
PR deletes layers, hazardsets and corresponding output in DB that were not harvested from geonode (without using force), to prevent conflicts between thinkhazard application and geonode data

Notes:
* does not handle changing hazardset ids (eg. SSBN => FATHOM) => should not happen, but can be resolved using `harvest --force`

### DIFF
--- a/docker/testdb/Dockerfile
+++ b/docker/testdb/Dockerfile
@@ -1,3 +1,3 @@
-FROM camptocamp/postgres:9.6
+FROM camptocamp/postgres:12
 
 COPY docker-entrypoint-initdb.d /docker-entrypoint-initdb.d
--- a/tests/processing/test_harvesting.py
+++ b/tests/processing/test_harvesting.py
@@ -317,6 +317,29 @@ class TestHarvesting(BaseTestCase):
         self.assertEqual(DBSession.query(Layer).count(), 1)
         self.assertEqual(DBSession.query(HazardSet).count(), 1)
 
+    @patch.object(Harvester, "fetch", side_effect=[
+        layers([{"id": 1}]),
+        layers([{"id": 1}, {"id": 2}]),
+    ])
+    @patch.object(
+        httplib2.Http,
+        "request",
+        side_effect=[
+            (Mock(status=200), json.dumps(layer({"id": 1, "hazard_period": 15}))),
+            (Mock(status=200), json.dumps(layer({"id": 1, "hazard_period": 15}))),
+            (Mock(status=200), json.dumps(layer({"id": 2, "hazard_period": 10}))),
+        ]
+    )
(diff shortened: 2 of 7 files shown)


thinkhazard/processing/harvesting.py
```diff
@@ -394,6 +398,26 @@ def harvest_layers(self, hazard_type=None):
...
+                logger.info("Deleting hazardset {} with output".format(hazardset.id))
+                self.dbsession.query(Output) \
+                    .filter(Output.hazardset_id == hazardset.id) \
+                    .delete()
+                self.dbsession.flush()
+                self.dbsession.delete(hazardset)
+            else:
+                logger.info("Keeping hazardset {}".format(hazardset.id))
+                hazardset.completed = False
+                hazardset.processed = None
```
arnaud-morvan, line 419: You should not iterate on hazardsets, but just alter hazardsets for not harvested layers.
Here you invalidate all processing outputs even if nothing has changed.
Here is the example I gave you in slack:
```
if layer.is_harvested() is False:
    session.delete(layer)
    hazardset = layer.hazardset
    session.query(Output).filter(Output.hazardset_id == hazardset.id).delete()
    session.flush()
    if len(hazardset.layers) == 0:
        session.delete(hazardset)
    else
        hazardset.completed = False
        hazardset.processed = None
```

tkohr: Ok thanks I moved it back into the layer iteration as discussed (just couldn't get pop to work). However the output deletion still doesn't seem correct.

tests/processing/test_harvesting.py
```diff
@@ -262,26 +262,49 @@ def test_scientific_quality_change(self, request_mock, fetch_mock):
...
+            layers([{"id": 1}, {"id": 2}]),
+            layers()
+        ]
+    )
+    @patch.object(
+        httplib2.Http,
+        "request",
+        side_effect=[
+            (Mock(status=200), json.dumps(layer({"id": 1}))),
+            (Mock(status=200), json.dumps(layer({"id": 2})))
```
arnaud-morvan, line 294: requests should be called 3 times

docker/testdb/Dockerfile
```diff
@@ -1,3 +1,3 @@
-FROM camptocamp/postgres:9.6
+FROM camptocamp/postgres:12
```
arnaud-morvan, line 1: This is a little uncomfortable as we cannot import new administrative divisions anymore, nor locally nor on OpenShift.

tkohr: That's true, will keep it as is for now to stay consistent with current Openshift cluster

PR #899, opened 2020-04-06
Gswbth 17 remove datasets fixes

arnaud-morvan:


tests/processing/test_harvesting.py
```diff
@@ -270,41 +275,85 @@ def test_scientific_quality_change(self, request_mock, fetch_mock):
...
         self.assertEqual(DBSession.query(Layer).count(), 0)
+        self.assertEqual(DBSession.query(HazardSet).count(), 0)
 
     @patch.object(
         Harvester,
         "fetch",
         side_effect=[
             layers([{"id": 1}, {"id": 2}]),
-            layers()
+            layers([{"id": 1}])
```
tkohr, line 293: Is this needed? Doesn't `layers()` create one layer with id 1?

tests/processing/test_harvesting.py
```diff
@@ -270,41 +275,85 @@ def test_scientific_quality_change(self, request_mock, fetch_mock):
...
+            (Mock(status=200), json.dumps(layer({"id": 2, "hazard_period": 10}))),
+            (Mock(status=200), json.dumps(layer({"id": 1, "hazard_period": 15}))),
+            (Mock(status=200), json.dumps(layer({"id": 2, "hazard_period": 10}))),
+        ]
+    )
+    def test_no_useless_process_invalidate(self, request_mock, fetch_mock):
+        """Multiple layers matching same level should not invalidate processing uselessly"""
         self.harvester().harvest_layers()
-        self.assertEqual(DBSession.query(Layer).count(), 0)
+        self.assertEqual(DBSession.query(Layer).count(), 1)
```
tkohr, line 343: Do we have 1 layer here and not 2, because one supersedes the other?

tests/processing/test_harvesting.py
arnaud-morvan: Yes, but more than it works: it works and it's easy to understand.

tests/processing/test_harvesting.py
arnaud-morvan: Yes, it is exactly what I want to test.

First harvesting:
- Layer1 match level HIG
- Layer2 match level HIG
=> we keep only Layer2

Second harvesting (nothing has changed on geonode side):
- Layer1 match level HIG
- Layer2 match level HIG
=> we should not have set hazadset.processed to None

PR #903, opened 2020-04-15
Import admin divs on PG12

arnaud-morvan:


thinkhazard/processing/import.py
```diff
@@ -110,7 +128,7 @@ def do_execute(self):
             )
         )
 
-    def import_shapefile(self, path):
+    def import_shapefile_ogr2ogr(self, path):
```
tonio, line 131: Do you keep it for reference to get back to it once it's compatible with PG12 ?

arnaud-morvan: Yes, shp2pgsql setup take 200 Mo.
And Debian Sid has fixed GDAL 3.

arnaud-morvan:
@tonio : I didn't manage to install a precompiled fixed GDAL.
I found easier to install PostGIS from distrib than to compile GDAL.

tonio:
> I found easier to install PostGIS from distrib than to compile GDAL.

Sure! Good to merge then.

PR #907, opened 2020-04-21
Fix for openshift

arnaud-morvan:
See commits fo details


.github/workflows/ci.yaml
```diff
@@ -30,6 +30,7 @@ jobs:
           && (
             startsWith(github.ref, 'refs/tags/')
             || github.ref == 'refs/heads/master'
+            || github.ref == 'refs/heads/fix_for_openshift'
```
tkohr, line 33: This should be removed before merge I guess

thinkhazard/processing/harvesting.py
```diff
@@ -111,24 +110,6 @@ def argument_parser():
     def do_execute(self, hazard_type=None, use_cache=False):
         self.use_cache = use_cache
 
-        setting_path = os.path.join(
```
tkohr, line 114: :+1: don't force automatically

thinkhazard/processing/import.py
```diff
@@ -96,22 +96,18 @@ def do_execute(self, use_cache=False):
...
                 self.download_zip(level, zip_path)
 
-            shp_path = os.path.join(self.TMP_PATH, "adm{}_th.shp".format(level))
-            if not use_cache or not os.path.isfile(shp_path):
-                LOG.info("Decompressing data for level {}".format(level))
-                subprocess.run(["unzip", "-o", zip_path, "-d", self.TMP_PATH], check=True)
+            LOG.info("Decompressing data for level {}".format(level))
+            subprocess.run(["unzip", "-o", zip_path, "-d", self.TMP_PATH], check=True)
 
+            LOG.info("Importing data for level {}".format(level))
```
tkohr, line 102: :+1: yes should also be executed when using cache

.dockerignore
```diff
@@ -1,7 +1,20 @@
-docker
-thinkhazard/static/build
-thinkhazard/static/font
-thinkhazard/locale/**/*.mo
-thinkhazard/locale/**/*.po
-thinkhazard.egg-info
-Dockerfile
```
tonio, line 7: Looks like it was shorter with blacklist than with whitelist ?

arnaud-morvan: Yes, but black list was not exhaustive (.covegare, .nodeids, eclipse files, ...).

Until now I had to permanently keep some custom files and folders here, for example "tmp" that I mount as volumes for caches and raise errors on build (owner postgres).

And I do not want to put custom .mk files with access keys and password in docker images that I could publish on docker hub in case of manual push.

Now everybody should build approximately the same locally as on CI.

arnaud-morvan: => Better reproduce what happens on CI

PR #910, opened 2020-04-27
[GSWBTH-47] Harvests hazard_unit for preprocessed layers

tkohr:
... but keeps logic to not harvest prepocessed layers with hazard_period which ensures to prevent harvesting multiple preprocessed layers for same hazardset

### DIFF
--- a/thinkhazard/processing/harvesting.py
+++ b/thinkhazard/processing/harvesting.py
@@ -584,11 +584,12 @@ class Harvester(BaseProcessor):
         mask = False
         if preprocessed is True:
             hazardlevel = None
-            hazard_unit = None
-            if o['hazard_period']:
-                logger.info('  return period found in preprocessed hazardset')
-                return False
-            hazard_period = None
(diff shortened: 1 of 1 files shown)


thinkhazard/processing/harvesting.py
```diff
@@ -584,11 +584,12 @@ def harvest_layer(self, object):
         mask = False
         if preprocessed is True:
             hazardlevel = None
-            hazard_unit = None
-            if o['hazard_period']:
-                logger.info('  return period found in preprocessed hazardset')
-                return False
-            hazard_period = None
+            # keep hazard_unit and hazard_period optinal for preprocessed layers
```
arnaud-morvan, line 587: optinal => optional

tkohr:
@arnaud-morvan I re-established the logic for preprocessed layer hazard_period after noticing its necessity and today's discussion

arnaud-morvan:
Ok, looks good, do what you think is better, merge or wait for discussion.

PR #913, opened 2020-04-30
[GSWBTH-38] Refactors S3Helper

tkohr:


thinkhazard/scripts/s3helper.py
```diff
@@ -39,3 +43,24 @@ def download_file(self, object_name, file_name=None):
...
+            self.s3_client.delete_object(
+                Bucket=self.bucket,
+                Key=object_name
+            )
+        except ClientError as e:
+            logging.error(e)
+            return False
+        return True
+
+    def check_object_exists(self, object_name):
```
tkohr, line 58: @arnaud-morvan Should this be implemented differently to not always provoke an exception in tween if maintenance file is not present in s3?

thinkhazard/views/pdf.py
```diff
@@ -151,7 +151,7 @@ def create_pdf_report(request):
         tempfile.gettempdir(), object_name
     )
 
-    s3_helper = _create_s3_helper(request)
+    s3_helper = S3Helper(request.registry.settings)
```
tonio, line 154: Have you checked performance cost when hitting real S3 backend at each request?

Plus, if you create the helper here, you'd better store in the request for reuse everywhere instead of creating other helpers in the code.

thinkhazard/scripts/s3helper.py
```diff
@@ -1,20 +1,24 @@
...
 class S3Helper:
-    def __init__(self, bucket, **kwargs):
-        # only set endpoint_url for debug, boto generates aws endpoint_url for prod
-        if kwargs["aws_access_key_id"] == "minioadmin":
-            kwargs["endpoint_url"] = "http://minio:9000/"
+    def __init__(self, settings):
+        # only set endpoint_url for dev, boto generates aws endpoint_url for prod
+        kwargs = ({"endpoint_url": "http://minio:9000/"}
+                  if os.environ['INI_FILE'] == "c2c://development.ini"
+                  else {})
```
arnaud-morvan, line 13: This is not a strong strategy.
We might use development.ini on OpenShift or production.ini locally.
I would recommend an AWS_ENDPOINT_URL env var.

thinkhazard/scripts/s3helper.py
arnaud-morvan: Seems good, if there is no more reliable way to get the info.

thinkhazard/views/pdf.py
arnaud-morvan: Or add this in main:
```
config.add_request_method(
    "s3_helper",
    lambda r: r.S3Helper(request.registry.settings),
    property=True,
    reify=True,
)
```
After what you can use ```request.s3_helper```

tkohr: Good points, thanks. I adapted as you proposed. 

Indeed the performance could be better (turning around 300ms for the remote s3 calls) but the s3 seems to be the most persistent solution in the new infra for now.

thinkhazard/scripts/s3helper.py
tkohr: True, adapted accordingly. Thanks

thinkhazard/views/pdf.py
arnaud-morvan: 300ms is reponse time when running the webapp locally, but actually home page request take 100ms on production server.

arnaud-morvan: We may use a table in public schema of public database ?

arnaud-morvan: Or it may be possible to completely restore the public database in a single transaction.
It such case, no need to disable public UI.

tkohr: Ok after discussion we take out maintenance mode for now due to short downtime and just take the refactoring part from this PR

PR #918, opened 2020-05-06
Always download in force mode

arnaud-morvan:
In Openshift container, download status in database is not reliable.


arnaud-morvan:
We may suppress the field from database and check the existance of file, but on OpenShift all task will be run in sequence from admin interface, so this does not really matter.

PR #923, opened 2020-05-13
[GSWBTH-51] Adds transifex tasks to admin interface

tkohr:
TODO: add tx env vars to openshift

Note: currently celery doesn't seem to work well with minio s3 backend in dev => sending celery logs to S3 removed for now


thinkhazard/celery.py
```diff
@@ -32,11 +31,17 @@ def publish():
 @app.task
 def transifex_fetch():
     print("start transifex_fetch")
-    # FIXME plug to real task
-    time.sleep(10)
+    os.system("/app/thinkhazard/scripts/tx-pull-db")
```
tonio, line 34: Does the output ends up the container log?

thinkhazard/templates/admin/index.jinja2
```diff
@@ -80,7 +79,20 @@
...
             </p>
-            -->
+            <p>
+              Push data from ThinkHazard DB to transifex for translation.
+            </p>
+            <p class="text-right">
+              <a
+                class="btn btn-info"
+                href="{{request.route_url('admin_add_task', _query={'task': 'transifex_push'})}}"
+                {% if 'transifex_push' in running_keys %}
```
tonio, line 89: We talked about this with Arnaud & thought it would be better to disable every task when there's one running. May can be done in this PR.

tkohr: Ah good point, I will adapt that.

thinkhazard/celery.py
tkohr: I can find entries such as {"status": "SUCCESS", "result": null, "traceback": null, "children": [], "date_done": "2020-05-13T12:40:19.284408", "task_id": "3743b2b9-6a12-4a3d-98ec-53e856ef5feb"} on the s3 for the corresponding time.

arnaud-morvan: Done, note that os.sytem doc say to use subprocess methods instead.
I checked and output are done in realtime on stdout.
Note that I would prefer to remove S3 result backend completely and get everything on container stdout.

tkohr:
@arnaud-morvan Yes different branches for the translations seems a good idea. Should we introduce another var env to handle that ?

thinkhazard/celery.py
tkohr: Not sure if I understand right. Would you remove all celery logging to S3?

Regarding subprocess I was wondering which one to use. Also for my understanding, what's the advantage here?

thinkhazard/templates/admin/index.jinja2
```diff
@@ -56,7 +56,7 @@
               <a
                 class="btn btn-info"
                 href="{{request.route_url('admin_add_task', _query={'task': 'process'})}}"
-                {% if 'process' in running_keys %}
+                {% if running|length > 0 %}
```
tonio, line 59: `running_keys` is not needed anymore, could be removed in the view

thinkhazard/celery.py
arnaud-morvan: Note that logging is not not actually send to s3, only results, but we do not know about the command runned. I would prefer having the results at end of log.

arnaud-morvan: Concerning suprocess, the advantage here => we get the command output in container output.

arnaud-morvan:
Yes I suppose we should add a variable, ex: TX_BRANCH_PREFIX="test_".
This might be the default in docker image and only removed on production environment.

thinkhazard/templates/admin/index.jinja2
tkohr: :+1:

thinkhazard/celery.py
tkohr: Ok I removed sending celery logs to S3 for now, which also resolves our problem between celery and minio in dev

PR #924, opened 2020-05-13
Fix PDF on India / OpenShift

arnaud-morvan:


production.ini
```diff
@@ -73,6 +73,7 @@ use = egg:gunicorn#main
 host = 0.0.0.0
 port = 8080
 workers = 3
+timeout = 120
```
tkohr, line 76: So this was the key to make it work, I guess

arnaud-morvan: yes, and after that the same on OpenShift / HAProxy : https://github.com/camptocamp/private-geo-charts/commit/fc6b1aabe170bddb042fa6b429498560710e2f8f

PR #926, opened 2020-05-18
Add htpassword based authentication

arnaud-morvan:


tests/views/test_admin.py
```diff
@@ -104,3 +130,8 @@ def test_climate_rec_edit(self):
         )
         form["associations"] = [admindiv.id for admindiv in admindivs]
         form.submit(status=302)
+
+    def test_preview(self):
+        resp = self.testapp.get("/preview/", status=302)
+        assert resp.location == "http://localhost/preview/en/"
```
tkohr, line 136: I guess tests run on default port 80?

thinkhazard/__init__.py
```diff
@@ -212,38 +234,47 @@ def redirect_to_default_language(request):
...
-    def wrapper_factory(request):
-        lang = request.matchdict["lang"]
-        # determine if this is a supported lang and convert it to a locale,
-        # likely defaulting to your default language if the requested one is
-        # not supported by your app
-        if lang not in request.registry.settings["available_languages"].split():
+    def text(self):
+        return 'languages = %s' % (self.val,)
+
+    phash = text
```
tkohr, line 245: Where is phash used?

thinkhazard/security.py
```diff
@@ -0,0 +1,12 @@
+import os
+from passlib.apache import HtpasswdFile
+
+
+def groupfinder(username, password, request):
+    ht = HtpasswdFile()
+    ht.load_string(os.environ.get("HTPASSWORDS", ""))
+
+    if ht.check_password(username, password):
```
tkohr, line 9: I guess decrypting passwords takes place here?

tests/views/test_admin.py
arnaud-morvan: Tests only create a WSGI application not a webserver.
`testapp` create requests with appropriate environnement variables, send them to WSGI application which return responses.
But no, there is no port opened here.

thinkhazard/__init__.py
arnaud-morvan: I suppose `phash` is shown in some pyramid log or command line tool output.
See https://github.com/Pylons/pyramid/blob/master/src/pyramid/predicates.py

thinkhazard/security.py
arnaud-morvan: Yes, but passwords cannot be decrypted (it would be a severe security issue !).
Passwords are encrypted and compared to the already encrypted ones stored in htpassword file, or env var in our case.

thinkhazard/security.py
```diff
@@ -0,0 +1,12 @@
+import os
+from passlib.apache import HtpasswdFile
+
+
+def groupfinder(username, password, request):
+    ht = HtpasswdFile()
+    ht.load_string(os.environ.get("HTPASSWORDS", ""))
```
tonio, line 7: passlib doc is unclear to me, what format takes this method? could be pass multiple users?

arnaud-morvan: This method take unicode or bytes, like `f.read()`.
Yes could pass multiple passwords, env var contains the htpassword file content.

PR #927, opened 2020-06-11
Add env vars for geonode

arnaud-morvan:
Note that this contains a small factorisation to use the `Harvester.fetch` function to retrieve documents dans layers details by id.


arnaud-morvan:
Merging to deploy, but comments are still welcome.

PR #928, opened 2020-06-12
Fix master

arnaud-morvan:
See commits for details


arnaud-morvan:
Merging as this fix master

PR #929, opened 2020-06-12
Cut geometries around meridian to avoid artefacts on projection

arnaud-morvan:


arnaud-morvan:
Merging to deploy, but comments are still welcome.

PR #937, opened 2022-02-01
Replace Mapbox style by camptocamp one

arnaud-morvan:
It seems that at some time there was two superposed styles (one from "stufraser1" and one from "gsdpm").
The one from "gsdpm" does not work anymore so I do not know how it should looks like.

The new one from "camptocamp" looks very similar to "stufraser1", a little more rivers and city labels appear at the same scale.

### DIFF
--- a/thinkhazard/static/js/report.js
+++ b/thinkhazard/static/js/report.js
@@ -11,8 +11,7 @@
   // Main
   //
   var sources = [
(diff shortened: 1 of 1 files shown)


arnaud-morvan:
Hi @pzwsk 

Here is old base map from stufrazer1:

stufrazer1

Here is new base map from camptocamp: 

camptocamp

stufraser1:
For me that is good to use. @matamadio @pzwsk

pzwsk:
+1 thank you Arnaud

PR #990, opened 2025-12-19
Use upsert strategy

arnaud-morvan:
- Add field admindiv.gaul
- Do not use TRUNCATE CASCADE
- Use UPSERT strategy for admindivs
- Maintain contacts and recommandations relations using GAUL
- Add class docstring


arnaud-morvan:
@ronitjadhav
I handled all relations properly. Don't hesitate to contact me by chat if something is not clear or if it desserve some comments, explanations.

PR #1016, opened 2026-02-09
Pyramid still require pkg_resources

arnaud-morvan:
But pk_resources has been removed from setuptools in v82?

See: https://github.com/pypa/setuptools/pull/5007


ronitjadhav (approved):
Thanks for the fix. I also noticed this issue today morning.

